PR #2, opened 2026-02-08
Stateless LLM calls with conversation compaction

lainn9527:
## Summary

- **Replace unbounded `--resume` history with stateless LLM calls**: Every GM call now sends system prompt + recent 20 messages instead of replaying the entire conversation. This caps per-turn context at ~21K tokens regardless of game length (vs ~157K at 200+ turns with `--resume`).
- **Add rolling narrative recap via `compaction.py`**: Background LLM summarization generates a 500-800 char recap of compacted history, injected into the system prompt via `{narrative_recap}` placeholder. Meta-compaction kicks in when recap exceeds 8K chars.
- **Add async tag extraction (`_extract_tags_async`)**: GM no longer needs to output structured STATE/LORE/NPC/EVENT tags — a separate post-processing LLM call extracts structured data from the narrative text, with dedup against existing data.

## Key Changes

| File | Change |
|------|--------|
| `compaction.py` | **NEW** — recap load/save, compaction triggers, background LLM summarization, branch recap copying |
| `claude_bridge.py` | Removed `--resume` codepath; both `call_claude_gm()` and `call_claude_gm_stream()` are now stateless |
| `app.py` | `RECENT_MESSAGE_COUNT` 10→20, `{narrative_recap}` in system prompt, removed all session_id bookkeeping, compaction triggers after GM responses, `copy_recap_to_branch()` on all branch creation |
| `auto_play.py` | Same pattern: recap in system prompt + compaction trigger after each turn |
| `event_db.py` | Added `get_event_titles()` for dedup in async tag extraction |
| `prompts.py` | Removed unused `_TAG_EXTRACTION_PROMPT` |

## Context Budget (per turn)

| Component | Est. Tokens |
|-----------|-------------|
| System prompt (filled, with recap) | ~8,000 |
| Augmented injection (lore + events + NPC activities) | ~1,500 |
| Recent 20 messages | ~11,500 |
| **Total** | **~21,000** (constant) |

## Test plan

- [ ] Start server, send a message — verify no `--resume` in logs (should show `mode=stateless`)
- [ ] Verify `{narrative_recap}` appears in filled system prompt with fallback text
- [ ] Play 25+ turns, verify `conversation_recap.json` created in branch dir
- [ ] Check recap content is narrative-focused (not listing stats/items)
- [ ] Test edit/regen — new branch should inherit parent recap
- [ ] Test auto_play — verify compaction triggers during long runs
- [ ] Switch between gemini/claude_cli — both should get same assembled context

### DIFF
--- a/app.py
+++ b/app.py
@@ -34,7 +34,7 @@ from parser import parse_conversation, save_parsed
 from prompts import SYSTEM_PROMPT_TEMPLATE, build_system_prompt
 from compaction import (
     load_recap, save_recap, get_recap_text, should_compact, compact_async,
-    get_context_window, copy_recap_to_branch,
+    get_context_window, copy_recap_to_branch, RECENT_WINDOW as RECENT_MESSAGE_COUNT,
 )
 
 # ---------------------------------------------------------------------------
@@ -53,7 +53,6 @@ LEGACY_CHARACTER_STATE_PATH = os.path.join(DATA_DIR, "character_state.json")
 LEGACY_SUMMARY_PATH = os.path.join(DATA_DIR, "story_summary.txt")
 LEGACY_NEW_MESSAGES_PATH = os.path.join(DATA_DIR, "new_messages.json")
 
-RECENT_MESSAGE_COUNT = 20
 
 DEFAULT_CHARACTER_STATE = {
     "name": "Eddy",
@@ -717,9 +716,9 @@ def _extract_tags_async(story_id: str, branch_id: str, gm_text: str, msg_index: 
             )
 
         except json.JSONDecodeError as e:
-            log.info("    extract_tags: JSON parse failed (%s), skipping", e)
+            log.warning("    extract_tags: JSON parse failed (%s), skipping", e)
         except Exception as e:
-            log.info("    extract_tags: failed (%s), skipping", e)
+            log.exception("    extract_tags: failed, skipping")
 
     t = threading.Thread(target=_do_extract, daemon=True)
     t.start()
@@ -1918,6 +1917,12 @@ def api_branches_edit():
     delta.append(gm_msg)
     _save_json(_story_messages_path(story_id, branch_id), delta)
 
+    # Trigger compaction if due
+    recap = load_recap(story_id, branch_id)
+    if should_compact(recap, len(full_timeline) + 1):
+        tl = list(full_timeline) + [gm_msg]
+        compact_async(story_id, branch_id, tl)
+
     log.info("/api/branches/edit DONE   total=%.1fs", time.time() - t_start)
     return jsonify({
         "ok": True,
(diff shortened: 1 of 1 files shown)


lainn9527 (commented):
## PR Review: Stateless LLM calls with conversation compaction

Nice PR — the design is solid and the bounded context approach is a clear win over unbounded `--resume` history. Here are my findings:

---

### Bug: `full_timeline` mutation race in `api_send`

In the non-streaming `api_send` compaction trigger:

```python
if should_compact(recap, len(full_timeline) + 1):
    full_timeline.append(gm_msg)      # mutates in-place
    compact_async(story_id, branch_id, full_timeline)
```

`compact_async` captures `full_timeline` by reference (closure), and `_run_compaction` slices it later in a background thread. This is a race if anything touches the list before the thread reads it. The same pattern appears in `auto_play.py`.

The **streaming route** does it correctly with `tl = list(full_timeline) + [gm_msg]` (a copy). The non-streaming route should do the same:

```python
tl = list(full_timeline) + [gm_msg]
compact_async(story_id, branch_id, tl)
```

---

### Missing compaction trigger in edit/regen routes

`should_compact` + `compact_async` is only wired into `api_send` and `api_send_stream`, but not into `api_branches_edit`, `api_branches_edit_stream`, `api_branches_regenerate`, or `api_branches_regenerate_stream`. If a user edits/regens many times on a long branch, compaction will never fire for those interactions.

---

### `_extract_tags_async` swallows exceptions silently

```python
except Exception as e:
    log.info("    extract_tags: failed (%s), skipping", e)
```

`log.info` loses the stack trace. Should be `log.warning` or `log.exception` so failures are debuggable.

---

### Minor

- **Dead `session_id` parameter**: Both `call_claude_gm()` and `call_claude_gm_stream()` still accept `session_id` "for API compat" but every call site passes `None`. Since this is an internal API, consider removing the parameter entirely.
- **`RECENT_WINDOW` duplication**: `compaction.py` defines `RECENT_WINDOW = 20` separately from `app.py`'s `RECENT_MESSAGE_COUNT = 20`. If they ever diverge, compaction boundaries won't match the actual context window. Consider importing from one place.
- **`_compact_locks` unbounded growth**: Locks are created per `(story_id, branch_id)` but never cleaned up. Not urgent (Locks are tiny), but worth a comment.

---

### Positive

- Clean, consistent removal of session_id across all 8 route variants
- Thread-safe compaction with per-branch `blocking=False` lock skip
- Atomic file writes via `.tmp` + `os.replace()`
- Good dedup logic in `_extract_tags_async` (checks existing lore topics + event titles)
- `copy_recap_to_branch()` correctly called on all 7 branch creation paths

**Recommendation**: Fix the `full_timeline` mutation race and add compaction triggers to edit/regen routes, then this is good to merge.

lainn9527:
Thanks for the quick fixes! Verified `cb8efe8`:

- Compaction triggers in all 6 routes — confirmed all use `tl = list(full_timeline) + [gm_msg]` (safe copy)
- `log.warning` / `log.exception` for tag extraction errors — good
- `RECENT_WINDOW` single source of truth via import — clean

You're right on item #4 — my mistake, the original code already used a copy. LGTM.

PR #8, opened 2026-02-08
Fix character name hallucination in summaries

lainn9527:
## Summary
- Both `auto_summary.py` and `compaction.py` were generating summaries without telling the LLM the player character's name
- The LLM hallucinated Chinese names (林浩/林皓/林陽/林昊) instead of the actual name "Eddy" from `character_state.json`
- Added `_get_character_name()` helper to both modules that reads from `branches/<bid>/character_state.json`
- Injected character name into LLM prompts with explicit instructions to use it consistently

## Root Cause
The summarization prompts tagged messages as 【玩家】/【GM】 but never specified who the player character actually is. Since all NPCs had Chinese names (林毅, 蘇凜, 阿豪), the LLM assumed the player also had a Chinese name and hallucinated different variants each time.

## Test plan
- [ ] Run auto-play and verify `auto_play_summaries.json` consistently uses the correct player name
- [ ] Trigger conversation compaction and verify `conversation_recap.json` uses the correct name in 3rd person
- [ ] Verify fallback to "玩家" if `character_state.json` is missing or has no `name` field

### DIFF
--- a/auto_summary.py
+++ b/auto_summary.py
@@ -8,6 +8,8 @@ import threading
 import time
 from datetime import datetime, timezone
 
+from story_utils import get_character_name
+
 log = logging.getLogger("auto_play")
 
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
@@ -15,17 +17,6 @@ STORIES_DIR = os.path.join(BASE_DIR, "data", "stories")
 
 SUMMARY_INTERVAL = 5        # every N turns
 MIN_COOLDOWN_SECONDS = 30   # minimum gap between summary calls
-
-
-def _get_character_name(story_id: str, branch_id: str) -> str:
-    """Read player character name from branch character_state.json."""
-    path = os.path.join(STORIES_DIR, story_id, "branches", branch_id, "character_state.json")
-    try:
-        with open(path, "r", encoding="utf-8") as f:
-            state = json.load(f)
-        return state.get("name", "玩家")
-    except (FileNotFoundError, json.JSONDecodeError):
-        return "玩家"
 
 # Track last run time per (story_id, branch_id) to enforce cooldown
 _last_run: dict[tuple[str, str], float] = {}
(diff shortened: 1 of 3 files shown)


lainn9527 (commented):
## Review — Game Designer Perspective

### Bug Analysis: Spot On

The root cause is well identified. The summarization prompts tag messages as 【玩家】/【GM】 but never tell the LLM the player's actual name. Since all NPCs have Chinese names (林毅, 蘇凜, 阿豪), the LLM infers the player must also have one and hallucinates a different variant each time (林浩/林皓/林陽/林昊). Clean diagnosis.

### What's Good
- Fix is precise and minimal — injects the name exactly where needed
- Fallback to `"玩家"` handles edge cases (missing file, no `name` field)
- `compaction.py` uses `{character_name}` placeholder in the template string — clean
- Explicit "do not substitute other names" instruction is the right call for CJK-heavy context

### Issues / Suggestions

1. **`_META_COMPACT_PROMPT` is missing character_name injection** — When recap exceeds 8K chars, meta-compaction re-summarizes it via `_META_COMPACT_PROMPT`, but that prompt has no character name hint. If the LLM re-hallucinates a name during meta-compaction, it corrupts the recap going forward. Low probability since the input recap should already contain the correct name, but worth defending against.

2. **`_get_character_name()` is fully duplicated** — Identical function copy-pasted into both `auto_summary.py` and `compaction.py`. Understandable since both modules intentionally avoid importing from `app.py`, but worth noting for future maintenance. A shared `utils.py` or similar would be cleaner.

3. **Prompt placement in `auto_summary.py`** — The character name instruction sits near the top, before `## 遊戲記錄`. After processing a wall of NPC Chinese names in the game log, the LLM may "forget" the earlier instruction (recency bias). Consider moving the `重要：...不要用其他名字替代` line closer to the output format instruction at the end of the prompt.

4. **Existing bad data not cleaned up** — The already-generated `auto_play_summaries.json` and `conversation_recap.json` files still contain hallucinated names. Not a blocker — can be a follow-up task.

### Verdict

LGTM — approve with the minor suggestions above. #1 (meta-compact prompt) is the most impactful one to address.

lainn9527:
All 3 issues from the previous review are addressed:

1. **Dedup resolved** — `get_character_name()` extracted to `story_utils.py`, imported by both modules
2. **Meta-compact prompt fixed** — `_META_COMPACT_PROMPT` now includes `{character_name}` placeholder and `.format(character_name=...)` call
3. **Prompt placement improved** — character name instruction moved to `## 重要規則` section after the game log, leveraging recency bias

Clean and minimal. LGTM — ready to merge.

PR #14, opened 2026-02-08
PR 1: World Timer + Day/Night Display

lainn9527:
## Summary
- Add `world_timer.py`: per-branch world_day tracking with TIME tag extraction, thread-safe advancement, and dungeon phase costs
- Integrate into all branch creation routes (create/edit/regen/promote/merge) via `copy_world_day()`
- API responses (`/api/messages`, `/api/status`) include `world_day` field
- Frontend header shows `✦ 世界第 N 天 · 時段` (5 time periods: 深夜/清晨/上午/下午/夜晚)

- **P1**: Added `threading.Lock` per branch for `advance_world_day()` (race condition fix)
- **P1**: Used regex in `parse_time_tag()` instead of raw `float()` (trailing text crash fix)

## Files changed
| File | Change |
|------|--------|
| `world_timer.py` | **New** — world day tracking, TIME tag parsing, dungeon costs |
| `app.py` | Import + integrate: `_process_gm_response()`, `api_messages()`, `api_status()`, all branch creation routes |
| `static/app.js` | `updateWorldDayDisplay()` function, called after loadMessages and during live polling |
| `static/style.css` | `#world-day-display` styling (gold, right-aligned in header) |
| `templates/index.html` | `<div id="world-day-display">` in header |

## Test plan
- [ ] `python app.py` starts without errors
- [ ] Send message → API response includes `"world_day"` field
- [ ] If GM outputs `<!--TIME days:1 TIME-->`, world_day advances by 1
- [ ] Create branch → child inherits parent's world_day
- [ ] Header shows `✦ 世界第 1 天 · 上午` (or similar)
- [ ] Live polling updates world day display

### DIFF
--- a/app.py
+++ b/app.py
@@ -1604,7 +1604,7 @@ def api_status():
     """Return character state for a branch."""
     story_id = _active_story_id()
     branch_id = request.args.get("branch_id", "main")
-    state = _load_character_state(story_id, branch_id)
+    state = dict(_load_character_state(story_id, branch_id))
     state["world_day"] = get_world_day(story_id, branch_id)
     return jsonify(state)
 
--- a/static/app.js
+++ b/static/app.js
@@ -469,6 +469,7 @@ async function init() {
     updateInitStatus("載入角色狀態…");
     const status = await API.status(currentBranchId);
     renderCharacterStatus(status);
+    updateWorldDayDisplay(status.world_day);
 
     // 6. Load NPCs, events, and summaries
     updateInitStatus("載入角色與事件…");
@@ -2060,7 +2061,9 @@ async function sendMessage() {
           smartScrollToBottom();
         }
 
-        renderCharacterStatus(await API.status(currentBranchId));
+        const status = await API.status(currentBranchId);
+        renderCharacterStatus(status);
+        updateWorldDayDisplay(status.world_day);
         loadNpcs();
         loadEvents();
       },
--- a/static/style.css
+++ b/static/style.css
@@ -1656,4 +1656,7 @@ body {
   .summary-modal-body {
     padding: 14px 16px 16px;
   }
+
+  .header-sub { display: none; }
+  #world-day-display { font-size: 0.75rem; }
 }


lainn9527 (commented):
## UI/UX Review

### Looks Good
- **Visual hierarchy is well-considered.** Using `var(--gold)` at 0.78rem creates a secondary-level element that complements the header without competing with the main title (1.15rem bold) or subtitle (0.82rem dim). The gold color provides thematic consistency -- it already appears in dialogue and the sibling switcher label.
- **`:empty { display: none }` is the right call.** No phantom spacing when there is no world day data (e.g., fresh branch before first GM response). Clean and non-intrusive.
- **`white-space: nowrap` prevents awkward wrapping** of the day/period text within the element itself.
- **5 time periods are appropriate for the game design.** They map to natural language that a Chinese-language RPG audience recognizes instantly. Fewer (e.g., just day/night) would lose flavor; more (e.g., 8 three-hour blocks) would add noise without narrative value.
- **Fractional `world_day` encoding is clever.** Storing `Math.floor(worldDay)` as day count and `(worldDay % 1) * 24` as hour lets the frontend derive both day and period from a single float without extra API fields.

### Issues / Suggestions

**1. Mobile header overflow (P2)**
The `#header` is `display: flex` with no `flex-wrap`. On mobile (<=768px), the header already contains: hamburger (approx 40px) + title (approx 220px CJK chars) + subtitle (approx 120px) + gap spacing. Adding the world-day-display (approx 130px for `世界第 99 天 · 深夜`) pushes the total past 320px viewport width. During auto-play, the `#live-indicator` is also appended, making it worse.

**Suggestion:** Add a responsive rule in the `@media (max-width: 768px)` block. Options:
- Hide `.header-sub` on mobile and let `#world-day-display` take its visual slot (the subtitle is decorative; the world day is functional).
- Or hide `#world-day-display` on mobile and show the info elsewhere (e.g., in the character status panel which is always accessible via the drawer).
- Or reduce the title to a shorter form on mobile.

**2. Collision with `#live-indicator` margin-left: auto (P3)**
Both `#world-day-display` and `#live-indicator` use `margin-left: auto`. When both are rendered (during auto-play live view), `#world-day-display` claims the auto margin first (since it appears earlier in DOM order), and `#live-indicator` sits right next to it with no visual separation beyond the 12px flex gap. This works but:
- The two elements look like a single run-on text since both are small, non-bordered text elements.
- Consider adding a subtle left border or extra margin to `#live-indicator` when it coexists with the world-day display, or group them in a wrapper `<div>` with explicit spacing.

**3. Font size accessibility (P3)**
0.78rem is approximately 12.5px at default browser settings. This is below the generally recommended 14px minimum for body-adjacent UI text. On the dark background (`--bg-sidebar: #343330`) with gold (`#d4a853`), the contrast ratio is approximately 5.6:1, which passes WCAG AA for small text, so color is fine. But the size combined with `letter-spacing: 0.03em` may be hard to read for some users, especially on mobile.

**Suggestion:** Consider bumping to 0.82rem (matching `.header-sub`) or at least 0.8rem. It would still read as secondary to the title.

**4. Day count starts at 1 but could confuse with "Day 0" edge case (minor)**
The JS does `Math.floor(worldDay) + 1`, so `world_day: 0` shows "世界第 1 天". But the guard `if (worldDay <= 0) { el.textContent = ""; return; }` means `world_day: 0` hides the display entirely -- which is correct for "no data yet." However, `world_day: 0.0` (explicitly set) also hides. If the system intends 0.0 to mean "Day 1, midnight," this would be a bug. Worth clarifying the contract: is 0.0 "uninitialized" or "start of Day 1"?

### Verdict: APPROVE

The feature is well-integrated and the visual design fits the existing dark-theme aesthetic. The mobile overflow concern (point 1) is the most important item and should ideally be addressed before this ships to avoid a broken header on phones, but it is not a blocking issue if mobile is not a primary target right now. The other suggestions are minor polish items.

lainn9527 (commented):
## 🎮 Game Director Review

### ✅ Looks Good

- **Clean integration with existing tag pipeline**: TIME tag processing slots neatly into `_process_gm_response()` alongside STATE/LORE/NPC/EVENT/IMG tags. The pattern is consistent and the GM system prompt instructions (lines 87-96 of `system_prompt.txt`) are clear with good examples.
- **Thread-safety from day one**: Per-branch locks in `advance_world_day()` prevent race conditions with background NPC evolution and async tag extraction. Good foresight — this would have been a painful bug to debug later.
- **Branch inheritance is thorough**: `copy_world_day()` is called in all 7 branch creation paths (create, edit, edit-stream, regen, regen-stream, promote, merge). This matches the existing pattern for `copy_recap_to_branch()` exactly, so no branch will silently lose its clock.
- **GM-driven time with clear guardrails**: The system prompt instruction "日常對話不需要推進時間。只在敘事中明確有時間跳躍時使用" is the right boundary. It prevents clock inflation from casual conversation while letting the GM advance time during training montages, dungeon transitions, and rest periods.
- **Time period design fits the narrative**: The 5 periods (深夜/清晨/上午/下午/夜晚) map well to narrative beats — 深夜 for horror dungeon ambushes, 清晨 for waking up in 主神空間, 下午 for shopping at the exchange hall. The unequal spans (深夜 is 6h, 清晨 is 3h, 下午 is 6h) feel natural rather than mechanical.
- **Atomic file writes**: `_save_json` uses tmp + `os.replace()` for crash safety. Small detail, good practice.

### ⚠️ Issues / Suggestions

**1. Dungeon cost functions are dead code (low priority)**

`advance_dungeon_enter()` and `advance_dungeon_exit()` are defined in `world_timer.py` along with the `DUNGEON_TIME_COSTS` dict and a `"training": 2` cost, but none of these are imported or called from `app.py` or `auto_play.py`. The PR description mentions "Dungeon enter costs 3 days, dungeon exit costs 1 day" but this isn't actually wired up.

This isn't blocking — the GM can output `<!--TIME days:3 TIME-->` when narrating a dungeon transition, achieving the same effect. But the dead code creates a false promise: a reader sees `advance_dungeon_enter` and assumes it fires automatically on phase transitions. Either:
- Wire it into `auto_play.py`'s `update_phase()` where `dungeon_start`/`dungeon_end` is already detected, or
- Remove the functions and `DUNGEON_TIME_COSTS` dict, relying purely on GM `<!--TIME-->` tags.

I'd lean toward **wiring it in** — auto-play's phase detection (`_DUNGEON_START_PATTERNS` / `_DUNGEON_END_PATTERNS` regex in `auto_play.py`) already knows exactly when transitions happen. Having deterministic time costs for dungeon enter/exit (rather than hoping the GM remembers to emit a TIME tag) makes the clock more reliable, especially for multi-agent synchronization later.

**2. Time costs: 3 days enter / 1 day exit feels asymmetric — consider the narrative rhythm**

In the source material (無限恐怖), dungeon duration varies wildly (days to weeks inside the movie world). The question is whether `world_day` tracks *subjective time inside the dungeon* or *elapsed time in 主神空間's reference frame*.

- If **主神空間 time**: 3 days of "preparation/transit" before a dungeon makes sense narratively (the 主神 gives a countdown, team assembles, etc.). 1 day recovery after is fine.
- If **subjective total time**: 3 days is too short for most dungeons (Resident Evil alone spans multiple days in-world). The GM would need to add more TIME tags during the dungeon.

Suggestion: Add a brief comment in `world_timer.py` clarifying that `default_enter` is the *transit cost* (time consumed by the teleportation/preparation), and that in-dungeon time is tracked separately by GM TIME tags. This distinction matters for the multi-agent universe — if Agent A enters a dungeon and Agent B stays in 主神空間, they need to know whether 3 days pass for both or only for A.

**3. No TIME tag in system prompt's structured data injection — consider `{world_day}` placeholder**

The system prompt has `{character_state}`, `{narrative_recap}`, `{world_lore}`, `{npc_profiles}`. But there's no `{world_day}` placeholder. The GM has no way of knowing what the current world day/time period is unless it remembers from context.

This matters because:
- After conversation compaction, the recap may not mention the exact time.
- The GM might describe "morning sunlight" when it's actually 深夜.
- For multi-agent scenarios, agents need to know what time it is in other agents' timelines.

Suggestion: Inject current world day + time period into the system prompt (or as part of `{character_state}`). Even a one-liner like `當前時間：世界第 N 天 · 上午` would keep the GM grounded.

**4. No backward time movement — is this intentional?**

`advance_world_day()` silently ignores `days <= 0`. This means:
- No time reversal (fine — matches physics)
- No `<!--TIME days:-1 TIME-->` correction if the GM makes a mistake
- No `set_world_day` exposed via GM tags (only used internally for branch inheritance)

This is probably fine for now, but if you later have time-manipulation abilities (rewinding, temporal loops — very on-brand for 無限恐怖), you'll need to revisit this. Worth a TODO comment.

**5. Frontend display is minimal — consider tying time to visual atmosphere later**

The gold `✦ 世界第 N 天 · 深夜` header text is clean and unobtrusive. For a first iteration this is appropriate. But time periods become much more impactful if they affect the visual atmosphere:
- 深夜: darker background tint, muted colors
- 清晨: warm amber tones
- 夜晚: blue-tinted message area

This is purely a future suggestion — not needed for this PR. The data model supports it since `world_day` is already in the API response.

**6. Multi-agent synchronization: the per-branch model works, but needs a coordination layer later**

The PR description mentions this will support a multi-agent shared universe. The per-branch `world_day.json` model is the right foundation — each agent has their own clock, and branch operations (fork/promote/merge) propagate time correctly.

However, for true synchronization you'll need:
- A **global clock** or **reference timeline** that all agents' clocks can sync against
- **Time-gating**: Agent B can't advance past day 10 if Agent A is still on day 5 and they share 主神空間 (otherwise they could miss interactions)
- **Convergence points**: When agents return to 主神空間, their clocks should align

The current model doesn't prevent clock drift between agents, but it also doesn't make it harder to add sync later. The `set_world_day()` function is already there for forced alignment. This is fine to defer.

### 🎲 Balance Check: Time Period Distribution

| Period | Hours | Fraction of Day | Typical Activity |
|--------|-------|-----------------|-----------------|
| 深夜 | 0-6 (6h) | 25% | Sleep, ambushes, horror dungeons |
| 清晨 | 6-9 (3h) | 12.5% | Wake up, preparation |
| 上午 | 9-12 (3h) | 12.5% | Training, missions |
| 下午 | 12-18 (6h) | 25% | Main activity block, shopping, socializing |
| 夜晚 | 18-24 (6h) | 25% | Evening events, gambling, NPC encounters |

The distribution is reasonable. 清晨 and 上午 are short (3h each), which means they'll be "passed through" quickly during time skips. This is fine — mornings in 主神空間 are typically brief transition periods. If anything, you might later want to split 下午 into 下午 (12-15) and 傍晚 (15-18) for more granularity, but 5 periods is a good starting point.

### Verdict: APPROVE

The core design is sound — per-branch world day tracking, GM-driven time tags, and clean branch inheritance. The dead code for dungeon costs and the missing `{world_day}` system prompt injection are the main items to address in a follow-up, but neither blocks this PR. The foundation is solid for the multi-agent shared universe expansion.

lainn9527 (commented):
## Backend Review

### Looks Good

- **Thread-safety model is correct**: `_branch_locks` dict with a global `_branch_locks_lock` to guard creation, and per-branch `threading.Lock` for `advance_world_day()`. This is the right pattern for concurrent writes to per-branch files (e.g., NPC evolution background threads advancing time simultaneously with a user request).
- **Atomic writes via temp file + `os.replace()`**: `_save_json` in `world_timer.py` writes to `.tmp` then atomically renames. This avoids corrupt partial writes on crash. Well done.
- **Robust `parse_time_tag()` with regex extraction**: Using `_TIME_VALUE_RE = re.compile(r'[\d.]+')` to extract the numeric part is safe against trailing text, whitespace, or malformed GM output. The `try/except ValueError` on `float()` is a good second safety net.
- **Correct placement in `_process_gm_response()`**: `process_time_tags()` is called after IMG tag extraction and before `_extract_tags_async()`, which means: (a) TIME tags are stripped before the background LLM sees the text, and (b) the returned `gm_response` saved to messages is clean. Since `_process_gm_response()` is used by all 6 routes (send, send/stream, edit, edit/stream, regen, regen/stream) and by `auto_play.py`, this is a single integration point with full coverage.
- **Branch creation coverage is thorough**: `copy_world_day()` is added to all 7 branch creation/inheritance paths: create, edit, edit/stream, regenerate, regenerate/stream, promote, merge. This mirrors the existing `copy_recap_to_branch()` pattern exactly.
- **Blank branch correctly omitted**: `api_branches_blank()` does NOT call `copy_world_day()`, which is correct — blank branches are fresh starts (`branch_point_index: -1`) and should begin at day 0. `get_world_day()` returns 0 when the file is missing, so this works by default.
- **Graceful degradation**: `_load_json` returns `default` on `JSONDecodeError` or `OSError`, so a corrupt `world_day.json` simply resets to 0 rather than crashing.
- **API contract**: `world_day` is correctly exposed in both `/api/messages` (for `loadMessages` calls) and `/api/status` (for status polling).

### Issues / Suggestions

1. **[Minor] `api_status` mutates returned dict with `world_day` key** — `world_timer.py` is not involved here, but in `app.py` the PR does:
   ```python
   state = _load_character_state(story_id, branch_id)
   state["world_day"] = get_world_day(story_id, branch_id)
   return jsonify(state)
   ```
   `_load_character_state()` calls `_save_json(path, state)` before returning (line 246), so the in-memory mutation of `state["world_day"]` does not contaminate the persisted file. This is safe **today**, but it is fragile — if anyone later calls `_save_json` on this dict again, `world_day` would leak into `character_state.json`. Consider either: (a) `return jsonify({**state, "world_day": get_world_day(...)})` to avoid mutating the original, or (b) adding a comment explaining the mutation is intentional and read-only.

2. **[Minor] SSE streaming send `"done"` event does not include `world_day`** — The streaming send endpoint (`/api/send/stream`, line 1585-1589) returns `{"type": "done", "gm_msg": ..., "branch": ...}` but no `world_day` field. The frontend's `onDone` handler for send (app.js ~line 1991) does call `API.status()` which now includes `world_day`, but `renderCharacterStatus()` doesn't forward it to `updateWorldDayDisplay()`. This means after a normal send with a TIME tag, the header day/night display won't update until the next page refresh or branch switch. This is a frontend-side gap, but the fix likely involves either: (a) adding `world_day` to the SSE done payload and calling `updateWorldDayDisplay` in onDone, or (b) having the `onDone` handler call `updateWorldDayDisplay(data.world_day)` with the status response. Low severity since the data is consistent on the backend side and the display corrects on next `loadMessages`.

3. **[Nit] `_branch_locks` dict grows unboundedly** — Each unique `story_id/branch_id` pair creates a lock that is never removed, even after branch deletion. In practice this is negligible (a few hundred entries at most), but if this codebase ever supports many stories/branches, consider a cleanup hook or using `weakref`. Not a blocker.

4. **[Nit] `copy_world_day` skips day=0** — The guard `if day > 0` means if a parent branch is explicitly at day 0 with a `world_day.json` file, the child won't inherit the file. Since `get_world_day` defaults to 0 when missing, this is functionally correct but semantically surprising. A comment explaining the optimization ("no file needed for day 0") would help.

5. **[Nit] `parse_time_tag` silently ignores unknown units** — If the GM outputs `<!--TIME minutes:30 TIME-->`, it returns 0 and the time is lost. Consider logging a warning for unrecognized time units so it's easier to debug GM prompt issues.

### Verdict: APPROVE

The backend integration is clean and correct. The thread-safety model is sound, error handling degrades gracefully, all branch paths are covered, and the API contract is properly extended. The issues noted above are all minor/nit-level and none are blocking.

lainn9527:
## Fixes Applied (2ed85fa)

Thanks to all four reviewers. Here's what was addressed:

### BE Review
- ✅ **`api_status` in-place mutation**: Fixed — now uses `dict()` shallow copy before adding `world_day`
- ℹ️ **SSE streaming done event**: The `onDone` callback now calls `API.status()` → `updateWorldDayDisplay()`, so the display updates immediately after streaming completes
- ℹ️ **`_branch_locks` unbounded growth**: Acknowledged — negligible for expected branch counts (<1000). Will address if profiling shows issues
- ℹ️ **`copy_world_day` skips day=0**: Intentional — day 0 means "uninitialized/fresh start", no file needed
- ℹ️ **`parse_time_tag` unknown units**: Good point but keeping silent for now — unknown tags are no-ops by design

### FE Review
- ✅ **Missing `updateWorldDayDisplay` after `sendMessage()`**: Fixed — `onDone` callback now stores status and calls `updateWorldDayDisplay(status.world_day)`
- ✅ **`margin-left: auto` collision with `#live-indicator`**: Fixed — `#live-indicator` now uses `margin-left: 8px` (world-day-display pushes right via `margin-left: auto`, live-indicator follows with fixed gap)
- ✅ **Mobile styles**: Added `header-sub { display: none }` and `#world-day-display { font-size: 0.75rem }` in `@media (max-width: 768px)`
- ℹ️ **`worldDay = 0` hidden**: Intentional — 0 means no TIME tag received yet (game hasn't started tracking time)

### UI/UX Review
- ✅ **Font size**: Bumped from 0.78rem to 0.82rem (matches subtitle size)
- ✅ **Mobile header overflow**: Hides `.header-sub` on mobile and shrinks world-day font to 0.75rem
- ✅ **Live indicator collision**: Fixed with `margin-left: 8px`
- ℹ️ **Day 0 semantics**: 0 = uninitialized (no time tracking yet), not "start of Day 1"

### Game Director Review
- ℹ️ **`advance_dungeon_enter/exit` dead code**: These are wired up in PR #16 (auto-play agent integration) — called from `auto_play.py update_phase()` on dungeon phase transitions
- ℹ️ **Missing `{world_day}` in system prompt**: Good suggestion — will add as a follow-up. Currently the GM receives time context via `_build_augmented_message()` injection
- ℹ️ **Time semantics**: `world_day` is 主神空間 reference-frame time. Will add a docstring comment clarifying this
- ℹ️ **No backward time movement**: Intentional for now. Time manipulation could be a future mechanic

All downstream branches (PR #15, #16, #17) rebased onto the updated `feature/world-timer`.

PR #15, opened 2026-02-08
PR 2: Agent Manager + Shared World Backend

lainn9527:
## Summary
- Add `agent_manager.py`: agent CRUD and lifecycle management (create/start/pause/stop/delete) with thread-safe locking
- Add `shared_world.py`: snapshot-based cross-agent awareness, leaderboard, context injection for human players
- Add agent API routes: `GET/POST /api/agents`, `POST /api/agents/<id>/<action>`, `DELETE /api/agents/<id>`, `GET /api/leaderboard`
- Add `POST /api/agents/generate-character`: LLM-powered character generation from free text description
- Cross-agent context injected into `_build_augmented_message()` and `_build_story_system_prompt()`
- Agent branches filtered from `GET /api/branches` response

- **P1**: Added `_active_threads_lock` for thread-safe `_active_threads` dict access
- **P1**: `delete_agent()` waits for thread join BEFORE removing registry entry
- **P0**: `start_agent()` validates `agent_id` field exists before starting
- **P1**: `generate_snapshot_summaries()` holds snapshot lock during write
- `_find_mentioned_agent()` requires name >= 2 chars to avoid false positives

## Dependencies
- Depends on PR #14 (World Timer)

## Test plan
- [ ] `POST /api/agents/generate-character` with `{"prompt": "冷酷前軍人"}` → returns character JSON
- [ ] `POST /api/agents` with generated character → creates agent
- [ ] `GET /api/agents` → lists created agents
- [ ] `GET /api/branches` → agent branches NOT shown
- [ ] `_build_augmented_message()` includes `[其他輪迴者動態]` when snapshots exist
- [ ] `GET /api/leaderboard` returns data (empty if no snapshots yet)

### DIFF
--- a/agent_manager.py
+++ b/agent_manager.py
@@ -204,16 +204,20 @@ def start_agent(story_id: str, agent_id: str) -> bool:
     )
 
     def _run():
+        error_msg = None
         try:
             auto_play(config)
         except Exception as e:
             log.exception("Agent %s crashed: %s", agent_id, e)
+            error_msg = str(e)
         finally:
             with _get_agents_lock(story_id):
                 data = load_agents(story_id)
                 ag = data["agents"].get(agent_id)
                 if ag and ag["status"] == "running":
-                    ag["status"] = "stopped"
+                    ag["status"] = "error" if error_msg else "stopped"
+                    if error_msg:
+                        ag["error_message"] = error_msg
                     _save_agents(story_id, data)
             with _active_threads_lock:
                 _active_threads.pop(agent_id, None)
--- a/app.py
+++ b/app.py
@@ -2970,7 +2970,9 @@ def api_agents_action(agent_id, action):
         ok = agent_manager.stop_agent(story_id, agent_id)
     else:
         return jsonify({"ok": False, "error": f"unknown action: {action}"}), 400
-    return jsonify({"ok": ok})
+    if not ok:
+        return jsonify({"ok": False, "error": f"cannot {action} agent {agent_id}"}), 409
+    return jsonify({"ok": True})
 
 
 @app.route("/api/agents/<agent_id>", methods=["DELETE"])
@@ -2978,7 +2980,9 @@ def api_agents_delete(agent_id):
     """Delete an agent (branch data is kept)."""
     story_id = _active_story_id()
     ok = agent_manager.delete_agent(story_id, agent_id)
-    return jsonify({"ok": ok})
+    if not ok:
+        return jsonify({"ok": False, "error": "agent not found"}), 404
+    return jsonify({"ok": True})
 
 
 @app.route("/api/leaderboard")
--- a/shared_world.py
+++ b/shared_world.py
@@ -126,45 +126,52 @@ def generate_snapshot_summaries(story_id: str, agent_branch_id: str):
     """Generate narrative summaries for unsummarized snapshots.
 
     Called after auto_play completes or periodically during play.
-    Uses lock to prevent concurrent writes.
+    LLM calls are made outside the lock to avoid blocking save_agent_snapshot().
     """
     from llm_bridge import call_oneshot
 
+    # 1. Read snapshots and identify unsummarized indices (under lock)
+    path = _snapshots_path(story_id, agent_branch_id)
     with _get_snapshot_lock(story_id, agent_branch_id):
-        path = _snapshots_path(story_id, agent_branch_id)
         snapshots = _load_json(path, [])
-        if not snapshots:
-            return
+    if not snapshots:
+        return
 
-        updated = False
-        for snap in snapshots:
-            if snap.get("summary"):
-                continue
+    # 2. Generate summaries outside the lock
+    summaries: dict[int, str] = {}  # index → summary text
+    for i, snap in enumerate(snapshots):
+        if snap.get("summary"):
+            continue
 
-            cs = snap.get("character_state", {})
-            name = cs.get("name", "未知")
-            missions = snap.get("completed_missions", [])
-            phase = snap.get("phase", "unknown")
-            turn = snap.get("turn", 0)
+        cs = snap.get("character_state", {})
+        name = cs.get("name", "未知")
+        missions = snap.get("completed_missions", [])
+        phase = snap.get("phase", "unknown")
+        turn = snap.get("turn", 0)
 
-            prompt = (
-                f"用繁體中文寫一句話（50-80字）總結輪迴者「{name}」在回合{turn}的狀態。\n"
-                f"當前階段：{'副本中' if phase == 'dungeon' else '主神空間'}\n"
-                f"已完成副本：{', '.join(str(m) for m in missions) if missions else '無'}\n"
-                f"獎勵點：{cs.get('reward_points', 0)}\n"
-                f"狀態：{cs.get('current_status', '未知')}\n\n"
-                f"直接輸出摘要文字，不要加標記。"
-            )
+        prompt = (
+            f"用繁體中文寫一句話（50-80字）總結輪迴者「{name}」在回合{turn}的狀態。\n"
+            f"當前階段：{'副本中' if phase == 'dungeon' else '主神空間'}\n"
+            f"已完成副本：{', '.join(str(m) for m in missions) if missions else '無'}\n"
+            f"獎勵點：{cs.get('reward_points', 0)}\n"
+            f"狀態：{cs.get('current_status', '未知')}\n\n"
+            f"直接輸出摘要文字，不要加標記。"
+        )
 
-            try:
-                summary = call_oneshot(prompt)
-                if summary:
-                    snap["summary"] = summary.strip()
-                    updated = True
-            except Exception as e:
-                log.warning("snapshot summary generation failed: %s", e)
+        try:
+            summary = call_oneshot(prompt)
+            if summary:
+                summaries[i] = summary.strip()
+        except Exception as e:
+            log.warning("snapshot summary generation failed: %s", e)
 
-        if updated:
+    # 3. Write summaries back (under lock)
+    if summaries:
+        with _get_snapshot_lock(story_id, agent_branch_id):
+            snapshots = _load_json(path, [])
+            for i, text in summaries.items():
+                if i < len(snapshots) and not snapshots[i].get("summary"):
+                    snapshots[i]["summary"] = text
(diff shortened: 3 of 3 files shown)


lainn9527 (commented):
## Frontend Review (API Contract)

### Looks Good
- **Response shapes are consistent**: `GET /api/agents` returns `{"agents": [...]}`, `GET /api/leaderboard` returns `{"leaderboard": [...]}` — matches existing patterns like `GET /api/branches`.
- **Create agent returns the full agent object**: `{"ok": true, "agent": {...}}` — this is exactly what the frontend needs to update the UI without a second fetch.
- **Generate-character endpoint is well-designed**: The `{"ok": false, "error": "...", "raw": "..."}` response for JSON parse failures gives the frontend enough info to show a useful error message and even display the raw LLM output for debugging.
- **Agent branch filtering from `/api/branches`** is the right call — keeps the branch tree clean for the human player. The frontend (PR #17) handles viewing agent branches via direct `switchToBranch()` without needing them in the tree.
- **Hidden tag pipeline reuse** (`_process_gm_response`) ensures agents and human players share the same data extraction logic — no divergent behavior.
- **Validation on create**: Correctly rejects missing `name` (400) and missing `character_state` (400) with error messages.

### Issues / Suggestions

**1. Action endpoint (`POST /api/agents/<id>/<action>`) returns `{"ok": false}` with HTTP 200 and no error message**

When `start_agent()`, `pause_agent()`, or `stop_agent()` return `False`, the route returns:
```python
return jsonify({"ok": ok})  # 200 status, no "error" field
```
The frontend (PR #17) doesn't check the response at all:
```javascript
await API.agentAction(btn.dataset.id, "start");
loadAgents();  // just reloads regardless
```
This means the user gets silent failures. The frontend can't distinguish "agent not found" from "agent already running" from "agent missing id field". 

**Suggestion**: Return an `"error"` string and appropriate HTTP status when `ok` is False:
```python
if action == "start":
    ok = agent_manager.start_agent(story_id, agent_id)
    if not ok:
        return jsonify({"ok": False, "error": "agent not found or already running"}), 409
```
This matches the existing pattern throughout `app.py` where failures use `{"ok": False, "error": "..."}` with 400/404/409 status codes.

**2. Delete endpoint (`DELETE /api/agents/<id>`) — same issue: silent 200 on failure**

```python
ok = agent_manager.delete_agent(story_id, agent_id)
return jsonify({"ok": ok})  # No error message, 200 even if agent doesn't exist
```
**Suggestion**: Return 404 with error message when agent not found.

**3. Action endpoint should return the updated agent object**

Currently the frontend must call `loadAgents()` (which triggers `GET /api/agents`) after every action to see the updated status. This creates a race condition: the agent thread may not have updated `agents.json` by the time the GET arrives.

**Suggestion**: Return the updated agent in the action response:
```python
return jsonify({"ok": True, "agent": agent_manager.get_agent(story_id, agent_id)})
```

**4. `completed_missions` in leaderboard response is a count (int), but the field name implies an array**

In `shared_world.py`:
```python
"completed_missions": len(snapshot.get("completed_missions", [])),
```
The leaderboard returns an integer, but the field name `completed_missions` suggests a list. The agent object in `agents.json` also stores `completed_missions` as a list (in character_state). This naming collision will confuse frontend developers.

**Suggestion**: Rename to `completed_mission_count` in the leaderboard response, or return the actual list if the frontend needs mission names for display.

**5. `generate-character` returns HTTP 500 for LLM JSON parse failure — should be 422**

```python
except json.JSONDecodeError:
    return jsonify({"ok": False, "error": "LLM returned invalid JSON", "raw": raw}), 500
```
This is not a server error — it's an expected failure mode where the LLM output didn't conform to the expected schema. HTTP 422 (Unprocessable Entity) would be more appropriate and won't trigger server error monitoring/alerting.

**6. `POST /api/agents` doesn't accept `provider` in the request body**

The route reads `name`, `character_config`, and `auto_play_config` from the body, but `provider` is hardcoded to `"claude_cli"` in `create_agent()`'s default parameter. The agent object stores the provider field, and the frontend may want to let users choose (especially since the app supports both Gemini and Claude). Even if not needed now, accepting and passing through `provider` from the body would be trivial.

**7. Minor: `GET /api/leaderboard` calls `get_leaderboard(story_id)` without `world_day`**

The leaderboard API calls `get_leaderboard(story_id)` without passing `world_day`, which means it uses the latest snapshot for each agent. But the system prompt injection calls `get_leaderboard(story_id, get_world_day(story_id, branch_id))` which filters by world day. This means the leaderboard the frontend shows might differ from what the GM sees — could be confusing if agents are at different timeline points.

**Suggestion**: Accept an optional `?world_day=` query parameter for consistency, even if the default (latest) is fine for most cases.

### Verdict: APPROVE with suggestions

The API design is solid and the frontend (PR #17) can consume it as-is. The silent failures on action/delete endpoints (issues #1 and #2) are the most impactful — they'll make debugging agent state issues painful. The rest are quality-of-life improvements that can be addressed in follow-up PRs.

lainn9527 (commented):
## 🎯 UI/UX Review

### ✅ Looks Good

- **Character generation prompt is excellent**. The `CHARACTER_GEN_SYSTEM_PROMPT` asks for concrete, specific details — internal contradictions, speech patterns, secrets, a first-person opening message. This will produce characters that feel distinct and memorable rather than generic RPG archetypes. The 200-400 char personality range is a good sweet spot: enough depth for the GM to roleplay convincingly, short enough that users won't get a wall of text when reviewing a generated character.

- **Snapshot-based cross-agent awareness is the right call**. Reading from time-indexed snapshots rather than live branch data means the human player sees a consistent, non-flickering view of other agents. No race conditions leaking into the UX.

- **Agent branch hiding is correct by default**. Users managing their own story branches should not see `auto_<hex>` branches cluttering the drawer. The branch data is kept on delete, so there is a recovery path if needed later.

- **Name-mention detail expansion is a nice touch**. When the user types an agent's name, the GM gets a full profile injected — personality, stats, inventory, adventure summary. This means organic in-story references ("I want to talk to 林浩") trigger richer GM responses without the user needing to navigate to a separate screen. The 2-char minimum for name matching is a sensible guard against false positives with single-character names.

- **Leaderboard fields are well-chosen for the 主神空間 setting**. reward_points as primary sort, gene_lock level, completed missions count, and current phase all map directly to in-universe progression markers that players would naturally care about.

### ⚠️ Issues / Suggestions

1. **"paused" vs "stopped" distinction is unclear to users** (Medium priority)
   The status model has both `paused` and `stopped`, but from the code, both cause the agent thread to exit on the next turn check. The difference is only semantic: `paused` implies "will resume where it left off" while `stopped` implies "finished." But `start_agent()` treats both identically — it auto-detects saved state and resumes either way. If the frontend shows both as separate statuses with separate buttons (Pause / Stop), users will wonder what the difference is. **Suggestion**: Either (a) drop `stopped` and only use `paused` + a terminal `finished` state, or (b) make the distinction meaningful — e.g., `paused` auto-resumes on server restart while `stopped` does not. At minimum, add a comment in the API or data model documenting the intended UX difference so the frontend developer knows how to label these.

2. **Error state has no user-facing recovery path** (Medium priority)
   When an agent crashes, the `_run()` finally-block sets status to `stopped` (not `error`). The docstring mentions `error` as a possible state, and `start_agent()` allows resuming from `error`, but nothing in the current code actually sets `status = "error"`. This means either: (a) the error state is dead code and should be removed from the status model to avoid confusing the frontend, or (b) you intend to add error-setting logic later — in which case the API should return an `error_message` field so the frontend can show users *why* their agent stopped. A red "Error" badge with no explanation is a bad experience.

3. **Character generation has no preview/edit step in the API flow** (Low priority, frontend concern)
   The `/api/agents/generate-character` endpoint returns generated JSON, and `/api/agents` POST creates the agent. This two-step flow is good — it implies the frontend can show a preview. But the API provides no way to *regenerate with tweaks*. If a user gets a character they mostly like but want to adjust the personality, they must either (a) send a brand new prompt and lose the whole generation, or (b) manually edit the JSON before POST. **Suggestion**: Consider accepting an optional `base_character` param in the generate endpoint so the LLM can refine rather than regenerate from scratch. This is a "nice to have" — the current two-step flow is workable.

4. **`[其他輪迴者動態]` context could get noisy with many agents** (Low priority)
   The `get_agents_context()` function lists ALL agents, plus a top-5 leaderboard, plus potentially a full detailed profile if a name is mentioned. With 10+ agents, this could consume significant context window budget per turn. The leaderboard in the system prompt (`_build_story_system_prompt`) is already capped at top-3, which is good. **Suggestion**: Cap the status list in `get_agents_context()` to top-5 or top-8 agents by reward_points, with a note like "...還有N位輪迴者" for the rest. This keeps the injection concise.

5. **Leaderboard is missing "survival" flavor** (Low priority, suggestion)
   The 主神空間 setting is fundamentally about survival — characters die permanently. The leaderboard shows reward_points, missions, gene_lock, phase, and status, but nothing about *how dangerous their journey has been*. Consider adding `total_turns` (how long they've survived) or a `near_death_count` field in future iterations. This would make the leaderboard feel more thematic — "this agent survived 200 turns" tells a different story than "this agent has 5000 reward points."

6. **No agent "personality preview" in list endpoint** (Low priority)
   `GET /api/agents` returns the full agent object including the entire `character_config`. For a list/card view, the frontend would benefit from a short `tagline` or `short_description` field (e.g., "冷酷前軍人，渴望同伴卻害怕背叛") that can be shown on agent cards without parsing the full 200-400 char personality. **Suggestion**: Add a `tagline` field to the character generation prompt (20-30 chars), or let the frontend truncate. Not blocking, but would improve the agent list UX.

7. **`api_agents_action` returns `{"ok": false}` with no error detail** (Low priority)
   When `start_agent()` returns `False`, the API returns `{"ok": false}` with HTTP 200. The frontend has no way to distinguish "agent not found" from "agent already running" from "missing id field." **Suggestion**: Return distinct error messages and appropriate HTTP status codes (404 for not found, 409 for already running).

### Verdict: APPROVE

The core UX decisions are sound: two-step character generation, snapshot-based awareness, name-triggered detail expansion, and hidden agent branches. The character generation prompt is well-crafted and will produce interesting characters. The main actionable item is clarifying the paused/stopped distinction before the frontend is built, so the UI labels and buttons make sense to users. The other suggestions are improvements that can be addressed in follow-up PRs.

PR #16, opened 2026-02-08
PR 3: Auto-Play Agent Integration

lainn9527:
## Summary
- Add cooperative stop: `should_stop()` checks agent status in `agents.json` when `agent_id` is set
- Add world timer integration: `update_phase()` calls `advance_dungeon_enter()`/`advance_dungeon_exit()` on phase transitions
- Save agent snapshots on phase changes and every 20 turns via `shared_world.save_agent_snapshot()`
- Generate snapshot summaries in daemon thread when auto-play completes

- **P0**: Replace `sys.exit(1)` with `return` in resume failure (safe for background thread)
- **P1**: Fix summary index: use `full_tl[max(0, len(full_tl) - 20):]` instead of `full_tl[last_turn * 2:][-20:]`

## Dependencies
- Depends on PR #15 (Agent Manager + Shared World Backend)

## Test plan
- [ ] `agent_manager.start_agent()` → auto_play runs in background thread
- [ ] Agent saves snapshots on phase changes (visible in `agent_snapshots.json`)
- [ ] Pause agent via agents.json status → auto_play stops cooperatively
- [ ] world_day advances on dungeon enter/exit
- [ ] No `sys.exit` crash in background thread

### DIFF
--- a/auto_play.py
+++ b/auto_play.py
@@ -604,7 +604,8 @@ def analyze_response(
     return result
 
 
-def update_phase(state: RunState, analysis: dict):
+def update_phase(state: RunState, analysis: dict,
+                  story_id: str = "", branch_id: str = ""):
     """Update run state phase based on analysis results."""
     if analysis["death"]:
         state.death_detected = True
@@ -616,6 +617,9 @@ def update_phase(state: RunState, analysis: dict):
             state.dungeon_count += 1
             state.hub_turns = 0
             log.info(">>> Phase: hub -> dungeon (dungeon #%d)", state.dungeon_count)
+            if story_id and branch_id:
+                from world_timer import advance_dungeon_enter
+                advance_dungeon_enter(story_id, branch_id)
         else:
             state.hub_turns += 1
     elif state.phase == "dungeon":
@@ -625,6 +629,9 @@ def update_phase(state: RunState, analysis: dict):
             state.phase = "hub"
             state.hub_turns = 0
             log.info(">>> Phase: dungeon -> hub")
+            if story_id and branch_id:
+                from world_timer import advance_dungeon_exit
+                advance_dungeon_exit(story_id, branch_id)
 
 
 # ---------------------------------------------------------------------------
@@ -649,6 +656,14 @@ def should_stop(state: RunState, config: AutoPlayConfig) -> bool:
     if state.consecutive_errors >= config.max_errors:
         log.info("STOP: Too many consecutive errors (%d >= %d)", state.consecutive_errors, config.max_errors)
         return True
+    # Cooperative stop: check agent status in agents.json
+    if config.agent_id:
+        from agent_manager import load_agents
+        agents_data = load_agents(config.story_id)
+        agent = agents_data.get("agents", {}).get(config.agent_id)
+        if agent and agent.get("status") != "running":
+            log.info("STOP: Agent status changed to '%s'", agent.get("status"))
+            return True
     return False
 
 
@@ -747,7 +762,7 @@ def auto_play(config: AutoPlayConfig):
         state = load_run_state(story_id, branch_id)
         if state is None:
             log.error("No saved state found for branch %s", branch_id)
-            sys.exit(1)
+            return
         # Clear stale death flag (detection logic may have changed)
         state.death_detected = False
         log.info("Resuming from turn %d on branch %s", state.turn, branch_id)
@@ -830,7 +845,7 @@ def auto_play(config: AutoPlayConfig):
 
             # D. Analyze & update phase
             analysis = analyze_response(gm_response, story_id, branch_id)
-            update_phase(state, analysis)
+            update_phase(state, analysis, story_id, branch_id)
 
             # E. Log
             log_turn(story_id, branch_id, state, player_text, gm_response)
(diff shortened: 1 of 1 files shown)


lainn9527 (commented):
## 🎯 UI/UX Review

### ✅ Looks Good

1. **Cooperative stop via `agents.json` status check is clean.** The `should_stop()` addition reads agent status from disk each turn, which is simple and debuggable. File-based signaling avoids complex IPC and works well for the background-thread architecture.

2. **Phase-change snapshots are the right trigger points.** Hub-to-dungeon and dungeon-to-hub transitions are the most meaningful narrative moments (agent enters/exits a mission). Snapshotting here gives the leaderboard and `[其他輪迴者動態]` context the most interesting data to surface.

3. **World timer integration (advance_dungeon_enter/exit) is correctly scoped.** Only fires when `story_id` and `branch_id` are present, so standalone CLI usage is unaffected. The lazy imports keep the dependency lightweight.

4. **P0 fix: `sys.exit(1)` -> `return` is critical for background thread safety.** `sys.exit()` in a daemon thread would kill the entire Flask process. Good catch.

5. **P1 fix: Summary index `full_tl[max(0, len(full_tl) - 20):]`** is correct. The previous `full_tl[last_turn * 2:][-20:]` assumed each turn = 2 messages (user + GM), which breaks when GM responses get split, images are inserted, or system messages appear.

6. **Daemon thread for `generate_snapshot_summaries` on completion** -- correct choice. Summaries are non-critical enrichment data; blocking the auto-play exit path for LLM calls would be wrong.

---

### ⚠️ Issues / Suggestions

#### 1. Snapshot every 20 turns may be too sparse for leaderboard engagement

**Current**: Snapshots fire on phase changes + every 20 turns as a fallback.

**Problem**: If a typical dungeon run is 30-50 turns, a human player checking the leaderboard mid-session may see stale data (up to 20 turns old, which is ~60 seconds of agent play time). For a real-time "shared universe" feel, this can feel sluggish.

**Suggestion**: Consider reducing to every 10 turns, or making the interval configurable via `AutoPlayConfig` (e.g., `snapshot_interval: int = 20`). The snapshot write is cheap (append to a JSON array), so the I/O cost is negligible. If configurability is too much scope for this PR, at minimum add a `SNAPSHOT_INTERVAL = 20` constant at module level so it's easy to tune later.

**Severity**: Low -- functional but impacts perceived "liveness."

#### 2. 3-second cooperative stop latency needs a "stopping..." UI state

**Current**: User clicks pause -> `agents.json` status set to "paused" -> agent checks `should_stop()` at the start of next turn -> agent stops.

**Problem**: With `turn_delay=3.0` plus the actual LLM call time (5-15s for a GM response), the real latency from click to stop could be **3-18 seconds**. During this window, the agent status in `agents.json` says "paused" but the agent is still generating content. If the frontend polls agent status, it will show "paused" while new messages keep appearing -- a confusing mixed signal.

**Suggestion**: 
- The `agent_manager` should track a transitional state: when a running agent's status is changed to "paused" but the thread is still alive, the API response should indicate `"status": "stopping"` (or return a separate `"thread_alive": true` field).
- The frontend can then show a "Stopping..." spinner instead of immediately showing the paused state.
- This is a PR #15 concern more than #16, but the cooperative stop in this PR surfaces the issue.

**Severity**: Medium -- will cause user confusion when pause doesn't appear immediate.

#### 3. Summary generation after completion creates a "hollow" initial experience

**Current**: `generate_snapshot_summaries()` runs in a daemon thread **after** auto-play completes. Before that, all snapshots have `"summary": ""`.

**Problem**: If a human player checks `[其他輪迴者動態]` or the leaderboard while an agent is still running, the snapshot data exists but has no narrative summary. The `_build_profile_from_snapshot()` function in `shared_world.py` shows "冒險經歷：" only when `summary` is truthy, so it will simply be absent. This means:
- While an agent is running (the most interesting time to observe it), the context injection has no narrative flavor -- just dry stats.
- After the agent finishes and summaries generate, the data suddenly becomes rich. This transition could feel inconsistent.

**Suggestion**: Consider generating summaries incrementally -- e.g., call `generate_snapshot_summaries()` on phase changes (alongside the snapshot save), not just on completion. Phase changes already create snapshots, and a single `call_oneshot()` for a 50-80 char summary is cheap (~1-2s). This would give running agents live narrative context. If you want to keep this PR minimal, at least add a `TODO` comment noting this as a follow-up improvement.

**Severity**: Medium -- impacts the quality of the shared-universe experience during the most engaging period (while agents are actively adventuring).

#### 4. Minor: `log.debug` for snapshot save failure may hide important errors

```python
except Exception as e:
    log.debug("Failed to save agent snapshot: %s", e)
```

If `save_agent_snapshot` fails (e.g., disk full, permission error), `log.debug` will be invisible in normal log levels. Since snapshots feed the leaderboard and cross-agent context, a persistent save failure degrades the user experience silently.

**Suggestion**: Use `log.warning` for the in-loop snapshot save failures and `log.error` for the final snapshot save failure (which also gates summary generation).

**Severity**: Low -- operational concern, not direct UX.

---

### Verdict: **APPROVE**

The core integration is solid -- cooperative stop, world timer hooks, and snapshot lifecycle are all correctly implemented. The P0 and P1 bug fixes are important. The suggestions above are improvements for engagement quality, not correctness blockers. The "stopping..." state (suggestion #2) and incremental summaries (suggestion #3) would be the highest-impact follow-ups for UX polish.

PR #17, opened 2026-02-08
Add agent frontend UI with LLM character builder

lainn9527:
## Summary
- **Agent panel** in drawer: list agents with status badges (待啟動/運行中/已暫停/已停止/錯誤) and controls (start/pause/stop/view/delete)
- **LLM character builder**: user types a character concept → calls `/api/agents/generate-character` → LLM generates a deep character with backstory, personality, contradictions → preview/edit → confirm creation
- **Leaderboard panel**: expandable ranking of agents by reward_points with gold/silver/bronze badges
- **Auto-polling**: 5s polling via `setTimeout` when agents are running, timer cleared on branch switch

## Changes
- `static/app.js` (+263 lines): Agent API methods, panel rendering, creation flow, leaderboard, auto-polling
- `static/style.css` (+167 lines): Agent cards, status colors, rank badges, creation modal, spinner
- `templates/index.html` (+39 lines): Agent section, leaderboard section, agent creation modal

## Depends on
- PR #15 (feature/multi-agent-backend)
- Independent of PR #16 (feature/auto-play-agent)

## Test plan
- [ ] Click "新增輪迴者" → modal opens with textarea for character concept
- [ ] Type concept (e.g. "冷酷的前軍人") → click "生成角色" → spinner shows → character preview appears
- [ ] Edit any field in preview → click "建立輪迴者" → agent created
- [ ] Agent appears in panel with "待啟動" badge
- [ ] Start/pause/stop controls update agent status
- [ ] Leaderboard shows agents ranked by reward_points
- [ ] Switch branch → agent polling timer cleared (no stale polling)
- [ ] Delete agent → removed from panel

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -1785,21 +1785,14 @@ async function loadAgents() {
     const agents = data.agents || [];
     renderAgentPanel(panel, agents);
 
-    // Auto-poll if any agent is running
+    // Auto-poll if any agent is running (single fetch per cycle via loadAgents)
     const hasRunning = agents.some(a => a.status === "running");
     if (hasRunning && !_agentPollTimer) {
-      _agentPollTimer = setTimeout(async function poll() {
-        await loadAgents();
-        const d2 = await API.agents();
-        if (d2.agents && d2.agents.some(a => a.status === "running")) {
-          _agentPollTimer = setTimeout(poll, 5000);
-        } else {
-          _agentPollTimer = null;
-        }
+      _agentPollTimer = setTimeout(function poll() {
+        loadAgents();  // loadAgents re-evaluates hasRunning and re-schedules if needed
       }, 5000);
-    } else if (!hasRunning && _agentPollTimer) {
-      clearTimeout(_agentPollTimer);
-      _agentPollTimer = null;
+    } else if (!hasRunning) {
+      if (_agentPollTimer) { clearTimeout(_agentPollTimer); _agentPollTimer = null; }
     }
 
     // Load leaderboard
@@ -1818,7 +1811,7 @@ function renderAgentPanel(panel, agents) {
   panel.innerHTML = agents.map(a => {
     const statusCls = AGENT_STATUS_COLORS[a.status] || "agent-status-stopped";
     const statusLabel = AGENT_STATUS_LABELS[a.status] || a.status;
-    const charName = a.character_config?.character_state?.name || a.name;
+    const charName = escapeHtml(a.character_config?.character_state?.name || a.name);
 
     let controls = "";
     if (a.status === "created" || a.status === "paused" || a.status === "stopped" || a.status === "error") {
@@ -1828,7 +1821,9 @@ function renderAgentPanel(panel, agents) {
       controls += `<button class="agent-ctrl-btn agent-pause-btn" data-id="${a.id}" title="暫停">⏸</button>`;
       controls += `<button class="agent-ctrl-btn agent-stop-btn" data-id="${a.id}" title="停止">⏹</button>`;
     }
-    controls += `<button class="agent-ctrl-btn agent-view-btn" data-id="${a.id}" data-branch="${a.branch_id}" title="查看">👁</button>`;
+    if (a.branch_id) {
+      controls += `<button class="agent-ctrl-btn agent-view-btn" data-id="${a.id}" data-branch="${a.branch_id}" title="查看">👁</button>`;
+    }
     controls += `<button class="agent-ctrl-btn agent-delete-btn" data-id="${a.id}" title="刪除">✕</button>`;
 
     return `<div class="agent-card">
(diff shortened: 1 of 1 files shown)


lainn9527 (commented):
## 🎮 Game Director Review

### ✅ Looks Good

**1. Character Creation Flow — Strong Foundation**
The free-text concept → LLM generation → editable preview → confirm pipeline is exactly right for this genre. The placeholder example ("一個總是笑嘻嘻但內心極度不信任他人的前詐欺師") teaches by showing, not telling — players immediately understand the depth of input expected. The editable preview (name, personality, opening, physique, spirit) gives just enough control without overwhelming.

**2. Agent Card Status System — Clean and Readable**
The five-state model (待啟動/運行中/已暫停/已停止/錯誤) with distinct color codes is intuitive. The control buttons correctly adapt to state — only showing relevant actions (e.g., pause/stop when running, start when created/paused/stopped). Good defensive UX.

**3. Leaderboard — Immediate Social Proof**
Gold/silver/bronze visual ranking with reward_points is a simple but effective motivator. Rendering it alongside the agent panel in the drawer makes it glanceable — players will see rankings every time they open the drawer, creating passive engagement.

**4. Polling Architecture — Sensible**
The 5s `setTimeout` chain (not `setInterval`) with auto-cleanup on branch switch and when no agents are running avoids the common pitfall of zombie timers. Clearing `_agentPollTimer` in `switchToBranch` is a good catch.

**5. Delete Confirmation — Correct Safety**
Using the existing `showConfirm()` with "分支資料保留" messaging is important — players need to know deleting an agent doesn't destroy the story branch they might want to revisit.

---

### ⚠️ Issues / Suggestions

**1. Double API Call in Polling Loop (Bug)**
```js
_agentPollTimer = setTimeout(async function poll() {
  await loadAgents();                    // ← calls API.agents() inside
  const d2 = await API.agents();         // ← calls API.agents() AGAIN
  if (d2.agents && d2.agents.some(a => a.status === "running")) {
```
`loadAgents()` already fetches agents and re-renders. The second `API.agents()` call is redundant — it doubles the request rate. Either use the result from `loadAgents()` (have it return agents), or just check the DOM/state after `loadAgents()` completes. This will cause unnecessary server load when multiple agents are running.

**2. Agent Card Shows Too Little — Missing "Story Hook"**
The card currently only shows: name + status + control buttons. For a player managing multiple agents in a shared universe, this is not enough to care about each agent or tell them apart at a glance. Suggestions:
- Show 1-line personality excerpt or the character concept the player typed
- Show current turn count or last activity timestamp ("第 23 回合" / "3 分鐘前")
- Show the agent's current dungeon/scenario name if available
- Even a tiny progress indicator (turns survived, dungeons completed) would make each card feel alive

**3. Leaderboard Only Shows Name + Points — Missing Context**
`reward_points` alone is an opaque number. Players will ask "why is this agent ranked higher?" Consider:
- Showing dungeons_completed or turns_survived as secondary info
- Adding a tooltip or expandable row with brief stats
- The leaderboard API likely returns more fields — surface at least one more metric

**4. "View" Button Switches Branch but Gives No Spectating Context**
Clicking the eye icon calls `switchToBranch(branch_id)` and closes the drawer. But the player lands on a regular message view with no indication that this is an agent's story. The existing auto-play live view has a `● LIVE` indicator and disabled input — does the agent branch also trigger those affordances? If not, the player might accidentally try to type into an agent's story. This should be verified.

**5. Preview Fields Are Not Complete — Missing Key Character Data**
The preview shows: name, personality, opening, physique, spirit. But the LLM-generated character (`res.character`) likely contains more fields (backstory, contradictions, catchphrases as mentioned in the PR description). These fields are silently dropped because `confirmBtn.onclick` only reads the five hardcoded `#preview-*` elements:
```js
const name = document.getElementById("preview-name")?.value || ch.name;
```
Either render all generated fields as editable (dynamically iterate `ch` keys), or at minimum pass the full `ch` object through and only override the five editable ones. Currently `characterConfig` is missing backstory/traits/contradictions.

**6. No Loading State on Agent Actions**
Start/pause/stop buttons call the API and then `loadAgents()`, but there's no visual feedback during the request. If the server takes 1-2s (e.g., starting an auto-play process), repeated clicks are possible. At minimum, disable the button during the request:
```js
btn.addEventListener("click", async () => {
  btn.disabled = true;
  await API.agentAction(btn.dataset.id, "start");
  loadAgents();
});
```

**7. No Error Handling on Agent Actions**
`API.agentAction`, `API.deleteAgent` responses are not checked for `res.ok`. A failed start (e.g., LLM quota exhausted) silently does nothing. Show a toast or alert on failure.

**8. XSS Risk in Agent Name Rendering**
Agent names come from LLM-generated content and are interpolated directly into innerHTML:
```js
`<span class="agent-name">${charName}</span>`
```
If the LLM generates a name containing `<script>` or HTML entities, this is injectable. Same applies to leaderboard `e.name`. Use `textContent` assignment or a sanitization helper. This is especially important because agent names are LLM-generated, not user-typed.

**9. Modal Lacks Keyboard UX**
- No Escape key to close the agent creation modal
- No focus management (textarea not auto-focused on open)
- Other modals in the app may have these behaviors — this one should be consistent

**10. Concept Textarea Has No Character Limit or Guidance**
Players could type a 2000-character essay as a "concept" and send it to the LLM. Consider:
- A `maxlength` attribute or visible character counter
- Hint text about expected length (e.g., "1-2 sentences describing the character concept")

---

### Missing Features — What Would Make This More Compelling

1. **Agent Activity Feed**: A small "last action" line on each card (e.g., "正在探索 咒怨副本..." or "已陣亡 — 第 47 回合") would make the agent panel feel alive without requiring players to click "View" on each one.

2. **Agent Comparison / History**: After agents die or complete scenarios, there's no way to compare their journeys. A mini-summary per agent (dungeons visited, key events, cause of death) would create replay value.

3. **Bulk Agent Creation**: For the "shared universe" vision, players may want 3-5 agents. A "quick create" flow or templates would reduce friction.

4. **Notification on Agent Events**: When an agent dies, completes a dungeon, or encounters a key event, a passive notification (badge number on drawer toggle, or toast) would pull players back to check.

---

### Verdict: REQUEST_CHANGES

The character creation flow and overall architecture are solid — this is heading in the right direction. However, the double API call in polling is a bug that should be fixed, the XSS risk in LLM-generated names needs sanitization, and the agent cards need more information density to make players actually care about their agents. The preview field data loss (backstory/contradictions not persisted) also undermines the value of the LLM character builder.

Priority fixes before merge:
1. Fix double `API.agents()` call in polling loop
2. Sanitize LLM-generated names in innerHTML
3. Pass full character data through on confirm (don't drop fields)
4. Add error handling on agent action API calls

PR #19, opened 2026-02-08
Fix branch switch flicker and scroll hijack during streaming

lainn9527:
## Summary
- **Smooth branch switching**: Add opacity fade transition (0.15s) when switching variants via `< 1/N >` sibling navigator, eliminating screen flicker
- **Respect user scroll during GM streaming**: Track scroll intent so `smartScrollToBottom` only auto-scrolls when user was at bottom and hasn't scrolled away — prevents force-scrolling back after stream ends

## Test plan
- [ ] Switch between sibling branches using `< 1/N >` arrows — should fade smoothly, no flicker
- [ ] During GM streaming, scroll up mid-stream — page should stay where you scrolled, not jump back after stream completes
- [ ] During GM streaming, stay at bottom — should continue auto-scrolling as before

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -821,7 +821,7 @@ async function switchToBranch(branchId, { scrollToIndex, preserveScroll, forcePr
 
   if (preserveScroll || forcePreserve) {
     savedScrollTop = container.scrollTop;
-    isAtBottom = (container.scrollTop + container.clientHeight >= container.scrollHeight - 50);
+    isAtBottom = (container.scrollTop + container.clientHeight >= container.scrollHeight - 100);
   }
 
   // Add fade-out transition for smooth variant switching


lainn9527 (commented):
**Backend review: No concerns**

This is a frontend-only change (static/app.js). No backend routes, API payloads, or data flows are affected. No new API calls introduced. The streaming SSE consumption pattern is unchanged.

**Scroll handler cleanup:** Verified all 5 exit paths from `sendMessage`'s streaming flow properly call `removeEventListener('scroll', scrollHandler)`:
1. `onDone` callback (line 1993)
2. `onError` callback — SSE "error" event (line 197 in `streamFromSSE` calls the passed `onError`)
3. `onError` callback — guard clause "連線中斷" (line 234 in `streamFromSSE` calls the passed `onError`)
4. Outer `catch` block (line 2036)

No memory leak risk — handler is added once per `sendMessage`, removed in all completion paths. The handler is passive (read-only scroll position check) and triggers no network requests.

**Two minor observations (non-blocking):**

1. If `API.switchBranch()` or `loadMessages()` throws after the opacity fade-out (line 829-830), the container stays at opacity 0.4 with no recovery path. Low risk since `switchToBranch` has no try/catch today either, but worth noting for a future hardening pass.

2. The `scrollToIndex` path (line 911-919) does not restore opacity/transition, so if both `preserveScroll` and `scrollToIndex` were somehow passed together, opacity would remain at 0.4. In practice these are mutually exclusive call patterns, so not a real issue.

No backend or data-flow risks. Would approve if this weren't my own PR.

lainn9527 (commented):
## Frontend Review: PR #19 — Branch switch transition & scroll tracking

**Verdict: Request Changes** (using --comment because GitHub prevents request-changes on own PRs)

---

### 1. [Bug] Branch Switch Opacity Fade — Missing error-path cleanup

If `API.switchBranch(branchId)` (line 833), `loadMessages()` (lines 842-844), `API.status()` (line 846), or any other `await` between the fade-out (line 830) and the fade-in paths throws an unhandled exception, the container will remain at `opacity: 0.4` permanently. The `switchToBranch` function has no `try/catch` wrapper.

The sibling switcher arrows (lines 1301, 1317) call `switchToBranch(..., { preserveScroll: true })` without any `.catch()` handler, so a network error during branch switch would leave the messages container semi-transparent with no recovery path.

**Suggestion**: Wrap the body of `switchToBranch` in `try/finally`, or at minimum ensure that after the fade-out is applied, any error resets opacity:

```js
// At the top of switchToBranch, after setting opacity:
try {
  // ... existing logic ...
} catch (err) {
  container.style.opacity = '';
  container.style.transition = '';
  throw err;
}
```

Alternatively, a CSS-class-based approach (`container.classList.add('switching')`) would be more robust and easier to clean up.

### 2. [Nit] Opacity cleanup duplication

The fade-in + transition removal pattern is duplicated 3 times (lines 876-880, 891-894, 902-905). Consider extracting into a helper:

```js
function fadeInContainer(container) {
  requestAnimationFrame(() => {
    container.style.opacity = '1';
    setTimeout(() => { container.style.transition = ''; }, 150);
  });
}
```

### 3. [Minor] Scroll threshold mismatch

The `wasAtBottom` check in `sendMessage` uses a **50px** threshold (line 1964):
```js
container.scrollHeight - container.scrollTop - container.clientHeight < 50
```

But `smartScrollToBottom` (line 2184) uses a **100px** threshold:
```js
const isNearBottom = container.scrollHeight - container.scrollTop - container.clientHeight < 100;
```

Edge case: if the user is 60-99px from bottom when they send a message, `wasAtBottom` is `false` (auto-scroll disabled), but `smartScrollToBottom` would have scrolled. The new behavior is stricter than the old. Consider aligning to 100px to match, or document why the tighter threshold is preferred.

### 4. [Info] Scroll handler logic — confirmed correct

The `scrollHandler` only sets `userScrolledDuringStream = true` when the user is NOT at bottom, and never resets it. This is correct — once the user scrolls away, auto-scroll stays disabled for the rest of the stream. Programmatic scrolls from `smartScrollToBottom` trigger the handler but put us at bottom, so `!currentlyAtBottom` stays false. No issue here.

### 5. [OK] `removeEventListener` coverage — complete

Called in all three exit paths:
- `onDone` callback (line 1993)
- `onError` callback (line 2030)
- `catch` block (line 2036)

`streamFromSSE` also has a guard for "stream ended without done/error" which calls `onError`. All paths covered.

### 6. [OK] `createBranchItem` indentation — verified clean

106 lines removed, 106 lines added. Pure whitespace change from 4-space to 2-space indentation. No logic changes.

### 7. [Minor] `wasAtBottom` is captured after forced scroll

At line 1964, `wasAtBottom` is captured **after** the GM bubble has been appended and `scrollToBottom()` called (lines 1959-1960). So `wasAtBottom` is almost always `true` because we just scrolled programmatically. The real work is done by `userScrolledDuringStream`.

This isn't a bug (behavior is correct), but `wasAtBottom` is misleading. If you want it to reflect the user's actual scroll position before the stream starts, capture it **before** line 1943 (before appending the GM bubble and calling scrollToBottom).

---

### Summary

| Finding | Severity |
|---------|----------|
| Missing opacity error-path cleanup | **Must fix** — can leave UI stuck at 40% opacity |
| Threshold mismatch (50px vs 100px) | Minor — behavioral edge case |
| `wasAtBottom` captured too late | Minor — works by accident, misleading |
| Opacity cleanup duplication | Nit |
| `removeEventListener` coverage | OK |
| `createBranchItem` indentation | OK |

**Blocking issue**: #1 (opacity stuck on error). Please add error handling around the fade transition in `switchToBranch`.

lainn9527:
**Addressing FE review feedback** (commit 116729f):

1. **[Fixed] Error-path opacity cleanup**: Wrapped all async calls in `switchToBranch` in `try/catch` — `fadeIn()` is called in `catch` to restore opacity before re-throwing.
2. **[Fixed] Scroll threshold mismatch**: Changed from 50px to 100px in both `wasAtBottom` and `scrollHandler` to match `smartScrollToBottom`.
3. **[Fixed] `wasAtBottom` captured too late**: Moved capture to **before** `$messages.appendChild(gmEl)` and `scrollToBottom()` so it reflects actual user scroll position.
4. **[Fixed] Duplicated fade-in code**: Extracted into `fadeIn()` helper, used in all 3 paths + error catch.

PR #20, opened 2026-02-08
Add branch indicator in header

lainn9527:
## Summary
- Show current branch name in the top-right corner of the header bar
- `updateBranchIndicator()` called on: init, branch switch, story switch, loadBranches
- Monospace styled `.branch-indicator` with subtle dark theme styling
- Also: add `llm_config.json` to `.gitignore`, update dev guidelines (4 reviewers)

## Test plan
- [ ] Verify branch name displays in header on page load
- [ ] Switch branches — indicator updates immediately
- [ ] Switch stories — indicator updates
- [ ] Create blank branch — indicator shows new branch name
- [ ] On `main` branch — indicator shows "main"


lainn9527 (commented):
## UI/UX Review: Branch Indicator in Header

### Overall
The feature concept is sound — showing the current branch name in the header gives users constant awareness of which timeline they're on, which is important for a branching-narrative RPG. However, there are several layout and accessibility issues that should be addressed before merging.

---

### 1. Layout conflict with `#live-indicator` (Bug)

Both `.branch-indicator` and `#live-indicator` use `margin-left: auto`, which pushes them to the right side of the flex header. When both are visible simultaneously (viewing a live auto-play branch), they will compete for the same space. The `#live-indicator` is dynamically appended to `#header` via `document.getElementById("header").appendChild(el)`, so it will appear *after* the branch indicator in DOM order.

The result: the branch indicator gets pushed to the middle area by the live indicator's `margin-left: auto`. This breaks the intended right-aligned positioning of the branch indicator.

**Suggestion**: Only the branch indicator should have `margin-left: auto`. The live indicator should drop `margin-left: auto` and instead sit next to the branch indicator with a small gap, or the branch indicator should be hidden/condensed when live indicator is active.

### 2. No text truncation for long branch names (Bug)

The `updateBranchIndicator()` function falls through to `$indicator.textContent = currentBranchId` for branches without a name. Branch IDs like `auto_a1b2c3d4` or UUIDs can be moderately long, and user-named branches could be even longer. Without `max-width`, `overflow: hidden`, `text-overflow: ellipsis`, and `white-space: nowrap`, a long branch name will:
- Push the header layout wider than the viewport on mobile
- Cause horizontal scrolling on narrow screens

**Suggestion**: Add these to `.branch-indicator`:
```css
max-width: 120px;       /* or 140px */
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

### 3. No mobile-specific styles

The `@media (max-width: 768px)` block has no overrides for the branch indicator. On mobile (where this app acts as a novel reader with serif fonts), the header already contains: hamburger button + title + subtitle. Adding a fourth element with `padding: 4px 12px` and `font-size: 0.75rem` may crowd the header. The monospace font (`Monaco`, `Menlo`) is also visually at odds with the serif-heavy mobile reading experience.

**Suggestion**: Consider hiding or shrinking the indicator on mobile. At minimum, add:
```css
@media (max-width: 768px) {
  .branch-indicator {
    max-width: 80px;
    padding: 2px 8px;
    font-size: 0.7rem;
  }
}
```

### 4. Accessibility: Contrast ratio is insufficient

The color combination is:
- Text: `var(--text-dim)` = `#9b9689` (muted brownish gray)
- Background: `rgba(255, 255, 255, 0.05)` over header background `#343330`

Computing the effective background: `rgba(255,255,255,0.05)` on `#343330` ≈ `#3d3c39`. The contrast ratio of `#9b9689` on `#3d3c39` is approximately **2.9:1**, which fails WCAG AA for normal text (requires 4.5:1) and even fails the large text threshold (3:1) at this font size (0.75rem/12px).

The indicator contains *functional information* (which branch the user is on), so it should be legible. Other dim-text elements in the app (subtitle) are less critical, but the branch name is actionable context.

**Suggestion**: Use `var(--text)` (`#e8e4df`) or at least a brighter variant. Even `var(--gold)` (`#d4a853`) would be more readable and thematically appropriate for an active-state indicator.

### 5. Minor: Monospace font choice

The `font-family: 'Monaco', 'Menlo', monospace` is a fine choice for branch IDs (which are code-like identifiers). However, when displaying user-assigned branch names (which could be Chinese text), monospace fonts may render CJK characters with awkward spacing. Consider adding a CJK fallback:
```css
font-family: 'Monaco', 'Menlo', monospace, "Noto Sans TC", sans-serif;
```

### 6. Minor: No `title` attribute for tooltip

When branch names are truncated (per suggestion #2), users lose the ability to read the full name. Consider adding a `title` attribute in `updateBranchIndicator()`:
```js
$indicator.textContent = branch.name;
$indicator.title = branch.name;
```

---

### Summary of required changes
| # | Issue | Severity |
|---|-------|----------|
| 1 | `margin-left: auto` conflict with `#live-indicator` | Bug — layout breaks during auto-play |
| 2 | No text truncation | Bug — causes overflow on mobile |
| 3 | No mobile responsive styles | UX — header crowding on mobile |
| 4 | Contrast ratio ~2.9:1 fails WCAG AA | Accessibility |
| 5 | Monospace CJK fallback | Minor |
| 6 | Missing `title` attribute for truncated names | Minor |

Items 1-4 should be addressed before merge. Items 5-6 are nice-to-haves.

lainn9527 (commented):
## Frontend Review — LGTM

Reviewed all call sites where `currentBranchId` is reassigned. Coverage is complete — every path that changes the active branch ultimately calls `updateBranchIndicator()`, either directly or through `switchToBranch()` / `loadBranches()`.

### Minor issues (non-blocking)

**1. Redundant calls to `updateBranchIndicator()`**

In `init()` (line 436) and `switchToStory()` (line 555), `updateBranchIndicator()` is called explicitly right after `loadBranches()`. But `loadBranches()` itself already calls `updateBranchIndicator()` at line 575. The double call is harmless (idempotent) but unnecessary. Consider removing the explicit calls in `init()` and `switchToStory()` since `loadBranches()` handles it.

**2. `$indicator` should be cached as a module-level constant**

Every other DOM element in this file is cached at module scope (lines 373–389: `$messages`, `$input`, `$sendBtn`, `$promoteBtn`, etc.). The branch indicator is the only element looked up via `document.getElementById()` inside its function body on every call. For consistency with the codebase pattern, cache it:

```js
const $branchIndicator = document.getElementById("branch-indicator");
```

and use `$branchIndicator` directly inside `updateBranchIndicator()`. This also removes the need for the null guard (the element is always in the DOM via `index.html`).

### What looks good

- **Fallback logic is correct**: The `main` check comes first (important since `branches["main"]` may not exist as a key), `branch && branch.name` guards against undefined entries, and raw `currentBranchId` is a sensible final fallback.
- **No missed call sites**: All 5 direct assignments to `currentBranchId` and all 10+ indirect changes via `switchToBranch()` are covered.
- **CSS**: `margin-left: auto` in a flex header correctly pushes the indicator to the right. Coexistence with the dynamically-appended `#live-indicator` (also `margin-left: auto`) works fine — the first auto-margin absorbs the space.
- **HTML placement**: After `.header-sub`, before the dynamically-appended live indicator — correct ordering.

lainn9527 (commented):
**Frontend Follow-up Review -- Approved** :white_check_mark:

Both suggestions from the previous review are addressed:

1. **`$branchIndicator` cached at module scope** -- Confirmed. Declared at line 392 as `const $branchIndicator = document.getElementById("branch-indicator")` alongside all other `$`-prefixed DOM constants (`$promoteBtn`, `$newBranchBtn`, etc.). The function references it directly with no per-call DOM lookup.

2. **Redundant `updateBranchIndicator()` calls removed** -- Confirmed. The function is called in exactly two places: inside `loadBranches()` (line 576) and inside `switchToBranch()` (line 863). Neither `init()` nor `switchToStory()` contain a direct call -- they rely on `loadBranches()` to handle the update, eliminating the double-call.

No remaining issues. LGTM.

PR #21, opened 2026-02-08
Fix branch tree depth: sibling detection + chain collapse

lainn9527:
## Summary
- **Backend**: Add `_resolve_sibling_parent()` helper that walks up the ancestor chain when `branch_point_index <= parent's branch_point_index`, making new branches siblings instead of children. Applied to all 5 branch creation routes.
- **Frontend**: Collapse linear chains in `buildBranchTree()` — only increase depth when parent has multiple children. Single-child chains display flat.
- **UX**: Add temporary "delete previous result" (🗑) button after regen, auto-dismisses after 30s or on next action.

## Test plan
- [ ] Edit a message at branch origin → new branch is sibling (same parent), not child
- [ ] Regen at branch origin → same sibling behavior
- [ ] Edit/regen deep in a branch → still correctly creates child branch
- [ ] Sibling switcher `< 1/N >` counts are correct with resolved siblings
- [ ] Existing deep linear chains display flat in drawer
- [ ] Timeline messages are correct when switching between branches
- [ ] Auto-play with blank branch still works
- [ ] Regen shows 🗑 button → click deletes old branch → 30s timeout auto-dismisses
- [ ] 🗑 button clears on: send message, start editing, regen again, switch branch

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -1238,9 +1238,13 @@ function showDeletePreviousBtn(oldBranchId) {
   btn.title = "刪除上一個結果";
   btn.addEventListener("click", async (e) => {
     e.stopPropagation();
-    await API.deleteBranch(oldBranchId);
-    await loadBranches();
-    await loadMessages();
+    try {
+      await API.deleteBranch(oldBranchId);
+      await loadBranches();
+      await loadMessages();
+    } catch (err) {
+      console.error("Failed to delete previous branch:", err);
+    }
     clearDeletePreviousBtn();
   });
 
--- a/static/style.css
+++ b/static/style.css
@@ -728,9 +728,10 @@ body {
 
 /* ---------- Delete Previous Button ---------- */
 .delete-prev-btn {
+  left: 40px !important;
   color: var(--accent);
   opacity: 0.7;
-  font-size: 0.75rem;
+  font-size: 0.85rem;
   transition: opacity 0.2s;
   animation: fadeInBtn 0.3s ease;
 }


lainn9527 (commented):
## UI/UX Review — PR #21

### 1. Delete-Previous Button (🗑) — Positioning Overlap (Bug)

Both the existing regen button (`↻`) and the new delete-previous button share `position: absolute; bottom: 8px; left: 0` via the `.msg-action-btn` base class. Since `showDeletePreviousBtn()` inserts the new button as a sibling DOM node right after the regen button:

```js
actionBtn.parentNode.insertBefore(btn, actionBtn.nextSibling);
```

...both buttons will render at the exact same absolute position and visually stack on top of each other. The delete button will cover the regen button.

**Fix**: The delete-prev-btn needs its own `left` offset (e.g. `left: 40px` or `left: calc(0px + <regen-btn-width> + 8px)`) so it sits next to the regen button rather than on top of it. Alternatively, consider using a non-absolute layout for the action button row (e.g. a flex container at the bottom of the message).

### 2. Delete-Previous Button — No Confirmation Dialog

The click handler immediately calls `API.deleteBranch(oldBranchId)` with no confirmation step:

```js
btn.addEventListener("click", async (e) => {
  e.stopPropagation();
  await API.deleteBranch(oldBranchId);
  ...
});
```

The existing branch deletion in the drawer uses `showConfirm()` before deleting. Deleting a branch is destructive and irreversible — a single mis-tap (especially on mobile) would lose an entire branch of content. The 30s auto-dismiss timeout is good for discoverability, but does not protect against accidental taps.

**Fix**: Add `if (!(await showConfirm("刪除上一個分支？"))) return;` before the `API.deleteBranch` call. This is consistent with the existing deletion pattern and adds minimal friction since the user just regenerated and is likely expecting the prompt.

### 3. Depth Collapsing — Linear Chains Lose Hierarchy Context

The change in `buildBranchTree()`:

```js
walk(branch.id, kids.length > 1 ? depth + 1 : depth);
```

This flattens single-child chains to the same depth. Combined with `renderBranchList()` which groups by depth-0, a linear chain like `main → A → B → C` would produce 4 separate root-level accordion groups. Users would see them as independent top-level branches with no visual indication of their parent-child relationship.

This is particularly problematic because:
- The accordion behavior (only one expanded at a time) means switching between them feels like switching between unrelated storylines
- The existing indentation was the *only* visual cue for branch parentage in the drawer

**Suggestion**: Instead of completely flattening, consider capping depth at a maximum (e.g. `Math.min(depth + 1, 3)`) so deep chains don't nest excessively but some hierarchy is preserved. Alternatively, keep the depth increment but compress the per-level indent (e.g. reduce from 16px to 10px per level).

### 4. CSS Animation — Looks Good

The `fadeInBtn` animation (0.3s ease, opacity 0→0.7) is smooth and consistent with the existing design language. The `live-pulse` and `bounce` animations in the codebase use similar easing. The 0.7 resting opacity makes the button visible but not dominant — good balance. No issues here.

### 5. 30s Timeout — Appropriate

30 seconds is a reasonable window. It gives users enough time to read the regenerated response and decide, while not cluttering the UI permanently. The auto-clear on other actions (send, edit, regen, switch branch) is thorough.

### 6. Emoji Consistency — Minor

The existing action buttons use text-based Unicode symbols (`✎` for edit, `↻` for regen, `✕` for delete). The new button uses a full emoji `🗑` (U+1F5D1) which renders as a colorful glyph on most platforms. This creates a visual mismatch — the other buttons are monochrome text characters and the trash can is a colored emoji.

**Suggestion**: Consider using `🗑️` with `font-family: inherit` and ensure it renders as text, or switch to a Unicode text symbol like `␡` or simply use the Chinese text `刪` to match the monochrome aesthetic. Alternatively, keep the emoji but acknowledge the inconsistency is minor.

### 7. Mobile/Responsive — Touch Target Concern

The `.delete-prev-btn` sets `font-size: 0.75rem` which is smaller than the base `.msg-action-btn` at `0.85rem` (and `0.9rem` on mobile). Given that this is a destructive action button, a smaller touch target on mobile is counterintuitive — if anything it should be the same size or larger to prevent mis-taps. Combined with the lack of confirmation dialog (issue #2), this is a real risk on mobile.

---

**Summary**: Issues #1 (overlap) and #2 (no confirmation) are bugs that should be fixed before merge. Issue #3 (depth flattening) could use reconsideration or at least deliberate acknowledgment. The rest are minor polish items.

lainn9527 (commented):
## Backend Review: LGTM (no actionable issues)

Reviewed the `_resolve_sibling_parent()` function and all 5 route integration points. The implementation is correct and well-structured.

### 1. `_resolve_sibling_parent()` correctness

**Edge cases verified:**
- **Parent is "main"**: Loop condition `current != "main"` immediately exits. Returns `"main"`. Correct.
- **Single-step chain**: If branch A forks from main at index 50, and user edits at index 50 on A (`branch_point_index = 49`), since `49 < 50`, the condition `branch_point_index <= parent_bp` holds, we walk up to main. New branch becomes sibling of A. Correct.
- **Deep chain (A -> B -> C)**: Walk-up continues through multiple ancestors until finding one whose `branch_point_index` is less than the requested fork point, or reaching "main". Correct.
- **Branch with `branch_point_index: None`** (e.g., root-like branches): The `if parent_bp is not None` guard correctly breaks the loop. No comparison against `None`.
- **The `<=` comparison**: `branch_point_index <= parent_bp` means forking at exactly the parent's own branch point creates a sibling, not a child. This is the correct semantic — editing the first message of a branch should create a sibling, not nest deeper.

### 2. All 5 routes updated

Confirmed `_resolve_sibling_parent` is called in all 5 branch-creation routes:
- `api_branches_create` (line 1663)
- `api_branches_edit` (line 1808)
- `api_branches_edit_stream` (line 1921)
- `api_branches_regenerate` (line 2045)
- `api_branches_regenerate_stream` (line 2150)

Other branch-mutation routes (`blank`, `promote`, `merge`, `delete`) correctly do not need this treatment.

### 3. Downstream call correctness

The resolved `parent_branch_id` is reassigned in-place before all downstream calls. Verified that `_find_state_at_index()`, `_find_npcs_at_index()`, `copy_recap_to_branch()`, `get_full_timeline()`, and `copy_world_day()` all receive the resolved value. Since the resolved parent is always a valid ancestor whose timeline contains the `branch_point_index`, all snapshot lookups will work correctly.

### 4. No infinite loop risk

The loop strictly moves `current` toward the root (via `parent_branch_id`), and terminates when reaching `"main"` or when `parent_branch_id` is `None`/missing (fallback to `"main"` via `or "main"`). Branch trees are acyclic by construction, so the loop is bounded by ancestor chain depth.

### 5. Thread safety

`_resolve_sibling_parent` is a pure read-only function on an in-memory dict loaded fresh per request. No mutations, no shared state. Consistent with the existing codebase pattern where each request loads its own copy of the tree. No new concerns.

### Minor note on the frontend

The `buildBranchTree` depth change (`kids.length > 1 ? depth + 1 : depth`) is a nice complement — it flattens single-child chains visually for any legacy data that already has deep nesting. The `showDeletePreviousBtn` in the regen flow is a good UX addition with proper cleanup on all state transitions (switchToBranch, startEditing, regenerateGmMessage, sendMessage, and the 30s auto-timeout).

No issues found. Clean implementation.

PR #22, opened 2026-02-08
Add lore console with CRUD and LLM chat

lainn9527:
## Summary
- New `/lore` page with two-panel layout for managing world lore entries
- Left panel: browsable lore list grouped by category, search, expand/collapse, edit modal with create/update/delete
- Right panel: LLM chat for discussing and proposing lore changes via `LORE_PROPOSE` structured tags
- Proposal cards with accept/reject workflow that batch-applies changes
- Mobile responsive with tab switching between panels

### New Files
- `templates/lore.html` — Page template with two-panel layout and edit modal
- `static/lore.js` — Lore list rendering, chat streaming, proposal cards
- `static/lore.css` — Dark theme styles reusing CSS vars from `style.css`

### Modified Files
- `app.py` — `/lore` page route + 6 new API routes (CRUD + chat stream + apply)
- `lore_db.py` — Added `delete_entry(story_id, topic)`
- `templates/index.html` — Navigation link (📖) in header
- `static/style.css` — Header lore link styling

## API Routes
| Method | Route | Purpose |
|--------|-------|---------|
| GET | `/lore` | Lore console page |
| GET | `/api/lore/all` | All entries + categories |
| POST | `/api/lore/entry` | Create entry |
| PUT | `/api/lore/entry` | Update entry (supports rename) |
| DELETE | `/api/lore/entry` | Delete entry |
| POST | `/api/lore/chat/stream` | SSE streaming chat with proposal parsing |
| POST | `/api/lore/apply` | Batch-apply accepted proposals |

## Test plan
- [x] All CRUD routes tested with Flask test client (create, read, update, delete)
- [x] Edge cases: empty topic, duplicate create, nonexistent entry, rename
- [x] Batch apply with add/edit/delete proposals
- [x] HTML template renders with all expected elements
- [ ] Manual browser test: left panel shows entries grouped by category
- [ ] Manual browser test: search filters entries
- [ ] Manual browser test: edit modal creates/updates/deletes entries
- [ ] Manual browser test: chat streams responses and shows proposal cards
- [ ] Manual browser test: accept proposal updates lore list
- [ ] Manual browser test: mobile tab switching works

### DIFF
--- a/static/lore.js
+++ b/static/lore.js
@@ -441,8 +441,9 @@
       });
     }
 
-    // Accept with confirmation for delete, error recovery
-    card.querySelector(".btn-accept").addEventListener("click", async () => {
+    // Accept with confirmation for delete — use event delegation so retry buttons work too
+    card.addEventListener("click", async (ev) => {
+      if (!ev.target.closest(".btn-accept")) return;
       if (action === "delete") {
         if (!confirm(`確定要刪除「${proposal.topic}」？`)) return;
       }


lainn9527 (commented):
## Game Director Review — Lore Console (PR #22)

### Overall Assessment

This is a genuinely useful tool for game masters. Having a dedicated lore management page outside the game session flow is the right design — it separates "author mode" from "play mode," which is exactly what a GM needs when maintaining world consistency across 41+ entries and 7 categories. The two-panel layout (browsable list + AI chat) maps well to the GM's mental model: "let me look at what I have, then discuss changes with an assistant."

### LLM System Prompt Analysis

The system prompt at line 2781-2796 is functional but lean to the point of being under-specified for this use case:

**What works:**
- Injecting the full lore corpus (all entries, grouped by category) gives the AI complete context to detect contradictions and suggest edits. This is the right call — unlike the game's RAG-lite approach (top-5 by bigram search), the lore console needs exhaustive context because the GM might ask about cross-category consistency.
- The proposal tag format (`LORE_PROPOSE`) is distinct from the game's `LORE` tag, preventing any confusion between the two systems. Good separation of concerns.
- "先討論再提案" (discuss before proposing) is the right guardrail — it prevents the AI from carpet-bombing the GM with unsolicited rewrites.

**What's missing or could be stronger:**

1. **No awareness of how lore is consumed during gameplay.** The system prompt doesn't mention that lore entries are searched via CJK bigram matching and injected into GM context as `[相關世界設定]`. This matters because the AI assistant should know to write lore entries that are *searchable* — e.g., using consistent terminology that players are likely to mention, including relevant keywords, keeping the `[tag: ...]` format. Without this context, the AI might propose beautifully written entries that perform poorly in RAG retrieval.

2. **No guidance on the `[tag: ...]` and `[source: ...]` inline metadata format.** The existing 41 entries use `[tag: 體系1/體系2]` and `[source: 參考來源]` conventions (parsed by `_TAG_RE` in `lore_db.py`, used for tag-based search). The system prompt should instruct the AI to preserve or add these tags when proposing edits/adds, otherwise new entries will lack searchable tags and the tag search API (`search_by_tags`) won't find them.

3. **No mention of the 7 existing categories.** The AI sees the grouped entries, but the prompt doesn't explicitly list valid categories (主神設定與規則, 體系, 商城, 副本世界觀, 場景, NPC, 故事追蹤). Without this, the AI might propose entries with novel category names that fragment the taxonomy. The category dropdown in the edit modal only shows existing categories, but the AI proposal flow bypasses the modal entirely — a proposed "add" with a new category string will create it silently.

4. **No entry length guidance.** Some existing entries are substantial (800+ chars with structured subsections), while others are brief. The AI should know the expected granularity — is one entry per mechanic (e.g., "基因鎖") the right level, or should it suggest splitting/merging?

### Proposal Workflow

**The add/edit/delete triad covers the essential CRUD**, and the accept/reject card UI is a clean interaction pattern. However:

5. **Missing "rename" action.** The manual edit modal supports renaming (via `new_topic`), but the AI proposal format doesn't have a rename equivalent. If a GM says "rename '已遇NPC' to '已遭遇NPC'," the AI has no clean way to express this — it would have to propose a delete + add, which loses the conceptual link. The `api_lore_apply` route doesn't handle rename either. This is a gap the GM will notice when trying to clean up topic names.

6. **No "merge" action.** With 41 entries, a common GM task is merging two related entries (e.g., combining two NPC entries or consolidating overlapping mechanics). The current workflow requires the GM to manually orchestrate: edit one entry to include both, then delete the other. The AI could propose this, but the lack of a structured "merge" action means it would need two separate proposals with no guaranteed ordering.

7. **Proposal content is truncated to 200 chars in the card preview** (`truncate(proposal.content, 200)` in `createProposalCard`). For an edit proposal on a lore entry that might be 800+ chars, the GM can only see the first 200 chars before deciding to accept. This undermines informed decision-making. Consider showing a expandable/collapsible full content view, or at minimum a diff against the current entry for edit proposals.

8. **No undo after accepting a proposal.** Once the GM clicks "採用," the `applyProposals` call fires immediately with no confirmation dialog and no way to revert. For delete proposals especially, this is a one-way door. The manual delete path at least has a `confirm()` dialog — the proposal path should have equivalent protection.

### Welcome Message

The welcome message (lines 44-49 in `lore.html`) is concise and sets appropriate expectations. The three bullet points correctly frame the tool's scope: querying, mutating, and consistency-checking. 

9. **One suggestion:** The welcome message doesn't mention that changes made here are immediately reflected in the game. A GM might not realize that accepting a proposal updates `world_lore.json` and the search index in real time, affecting the next player turn's context injection. Adding a brief note like "變更會即時同步到遊戲中" would set the right expectations about the live nature of edits.

### Integration with Existing Systems

10. **The chat sends the entire lore corpus in the system prompt every request.** With 41 entries (some 800+ chars each), this could be 20-30K tokens of system prompt per chat turn. This is fine for Gemini's 1M context, but for Claude CLI it's a significant chunk of the context window. If lore grows beyond ~100 entries, this approach won't scale. Not a blocker for launch, but worth noting. A future optimization could send only entries relevant to the conversation topic.

11. **No chat history persistence.** The `chatMessages` array lives in JS memory only. If the GM refreshes the page or navigates away, the entire conversation is lost. For a tool where the GM might spend 20+ minutes discussing and refining world lore, losing that context is painful. The GM might accept some proposals but want to return to the conversation later to accept others. This is a significant usability gap for the intended workflow.

### Summary

The lore console is a valuable addition to the GM toolkit. The core CRUD + AI chat + proposal workflow is sound. The main gaps are:
- **System prompt needs lore-specific guidance** (tag format, categories, searchability) — items 1-4 above
- **Proposal cards need full content visibility** — item 7
- **No undo/confirmation on proposal accept** — item 8
- **Chat history lost on page refresh** — item 11

Items 1-4 and 7-8 are the most actionable for this PR. Items 5-6 (rename/merge) and 10-11 (scaling/persistence) could be follow-up work.

lainn9527 (commented):
## FE Review (Round 2)

### Issues

- **[bug]** Retry button on proposal error is dead — no click handler. When `applyProposals` fails, the error-recovery path at `lore.js:458-465` replaces `actionsDiv.innerHTML` with a new `.btn-accept` button. But the event listener was bound to the *original* `.btn-accept` element which is now gone from the DOM. The new retry button is inert. Fix: use event delegation on the `card` element (e.g. `card.addEventListener('click', e => { if (e.target.matches('.btn-accept')) { ... } })`) or re-bind after replacing innerHTML.

- **[bug]** `collapsedCats` toggle logic is inverted for the default-collapsed state. When `allCollapsed = true`, a category is open if it is in `collapsedCats` (i.e. it is an "exception" to the default). Clicking a category header at `lore.js:240-244` toggles the DOM class AND toggles membership in `collapsedCats`. But on the next `renderLoreList` call (e.g. from search input), the render logic at line 192-194 reads: `isException = collapsedCats.has(cat); isOpen = allCollapsed ? isException : !isException`. This is correct on render. However, the click handler toggles the DOM *and* the set independently — if `renderLoreList` is called between clicks (e.g. search-as-you-type), the state is consistent. This is fine on closer inspection. No issue here after re-analysis — withdrawing this point.

- **[moderate]** Entry preview uses CSS adjacent sibling selector `.lore-entry.selected + .lore-entry-preview` (`lore.css:275`), but inside sub-groups the DOM structure is: `.lore-subgroup-entries > .lore-entry + .lore-entry-preview`. The selector still works because `.lore-entry` and `.lore-entry-preview` remain direct siblings inside `.lore-subgroup-entries`. This is correct, no issue.

- **[moderate]** `parseTopicPrefix` matches `:` (ASCII colon) as a delimiter (`lore.js:110`). This will false-positive on topics that contain URLs or time formats like `"SCP：收容失效 09:15 事件"` — the ASCII colon in `09:15` won't match first because `：` (full-width) is checked before `:`, but a topic like `"概述:基本規則"` with ASCII colon at a short prefix will be treated as a sub-group delimiter. This seems intentional for the data set, but worth noting if topics ever include colons in non-delimiter positions. Low risk given the CJK-heavy data.

- **[moderate]** The `expand/collapse` full content in proposal cards uses `contentDiv.innerHTML = expanded ? full : short` (`lore.js:439`). Both `full` and `short` are already `escapeHtml()`-processed, so there is no XSS risk here. Good.

- **[low]** Search input has no debounce (`lore.js:633-635`). Each keystroke triggers `renderLoreList` which does `$loreList.innerHTML = html` and re-binds all event handlers. With 41 entries this is fine. If the lore list grows significantly (200+), this could cause input lag. Not blocking.

- **[low]** `modal-body input` and `modal-body textarea` set `width: 100%` (`lore.css:508`) without `box-sizing: border-box`. If the parent does not set `box-sizing`, the padding (8px 10px) will cause these elements to overflow their container by ~20px. Check if `style.css` sets `*, *::before, *::after { box-sizing: border-box; }` globally.

### Looks Good

- XSS safety is solid: all user-controlled content goes through `escapeHtml()` before innerHTML insertion. The `escapeHtml` implementation via `div.textContent`/`div.innerHTML` is correct. Proposal cards, entry rendering, category headers all properly escaped.
- IIFE wrapping prevents global scope pollution.
- AbortController on SSE fetch with proper cleanup in `finishStreaming()`.
- `response.ok` check before calling `.getReader()` prevents unhandled errors on non-200 responses.
- Mobile responsive handling with `hidden-mobile` class applied in HTML to prevent flash of both panels on load.
- Chat history trimming to 40 messages prevents unbounded memory growth.
- Collapsed state preservation across re-renders via `collapsedCats` Set is well-implemented.
- The sub-group parsing with `parseTopicPrefix` and `buildSubGroups` is clean logic with proper sorting.
- Toggle-all button that clears exceptions on toggle is the right UX pattern.
- Escape key handler for modal.
- Proper error states on proposal accept (loading indicator, error message).
- Good use of CSS variables from the existing dark theme throughout.
- `word-break: break-word` fix from round 1 feedback.
- Edit button touch target increase to 36x36px minimum.

### Verdict: NEEDS CHANGES

One real bug: the retry button rendered after proposal-apply failure has no event listener, making error recovery non-functional. The rest is solid work.

lainn9527 (commented):
## Game Director Review (Round 2)

### Context for this review

The lore has grown from the documented "41+ entries" to **564 entries** across **11 categories** (including organic new categories like "道具" and "基本屬性" created by the auto-extraction pipeline). The dominant categories are 主神設定與規則 (247), 副本世界觀 (143), and 場景 (48). Many entries follow the `prefix：suffix` naming convention (e.g., "生化危機：浣熊市 — 醫院三樓東翼走廊地形與戰術環境"), which makes the sub-group feature in commit 3 genuinely essential rather than nice-to-have.

### Addressed from Round 1

The Round 2 commit addressed all four of my primary Round 1 items:
- System prompt now includes tag format guidance (`[tag:]`, `[source:]`), searchability tips, and explicit category list -- items 1-3. Good.
- Welcome message now mentions live sync ("變更會即時同步到遊戲中") -- item 9. Good.
- Proposal cards now have expandable full content view ("展開全文/收起") -- item 7. Good.
- Delete proposals now require `confirm()` dialog -- item 8. Good.

### Remaining Issues

- **[Medium] System prompt will blow up the context window at 564 entries.** My Round 1 item #10 noted this as a future concern at 41 entries. But the *actual* lore has 564 entries. The system prompt injects the entire corpus as formatted text -- at this scale, that is easily 80-100K tokens of system prompt per chat turn. For Gemini 2.5 Flash (1M context) this works. For Claude CLI, this will consume nearly the entire context window before the conversation even starts. The lore chat route currently uses `call_claude_gm_stream` which goes through the provider bridge, so the behavior depends on the configured provider. This is not a blocker for this PR -- the current user is on Gemini -- but the prompt should at minimum log or warn when the injected lore exceeds a threshold (say 50K chars). Alternatively, consider sending a TOC (like the game's `get_lore_toc()`) with an instruction to request full entries on demand, or only inject entries from the category the user is discussing.

- **[Low] New auto-extracted categories ("道具", "基本屬性", "副本背景", "世界設定") are not curated.** The system prompt now says "新增設定時請使用上方現有分類", and `cat_list` is dynamically built from the actual data. This means the AI assistant will see 11 categories including the organically-grown ones. This is correct behavior -- the AI should work with what exists. However, the lore console would benefit from being the place where a GM can *clean up* category drift. The current edit modal has a category dropdown populated only from existing categories with no "custom" option. If a GM wants to consolidate "副本背景" into "副本世界觀", they would need to manually edit each entry. This is a follow-up concern, not a PR blocker.

- **[Low] Sub-group feature is excellent for the actual data.** Commit 3 adds prefix-based sub-grouping with `：`, `:`, and ` — ` delimiters. Looking at the real data, this is critical: there are 60+ entries under "生化危機：浣熊市" alone. The sub-group collapse provides essential organization. The alphabetical sorting (`zh-Hant` locale) is also the right choice for CJK categories. No issues here -- calling it out as genuinely good design that maps well to how the auto-extraction pipeline names entries.

- **[Low] No rename/merge in proposal format (Round 1 items 5-6).** Still missing, but acceptable for launch. A GM can achieve rename by manually editing, and merge by asking the AI to combine content into one entry then deleting the other. The AI can guide this via the existing add/edit/delete triad.

### Looks Good

- **System prompt is now well-specified for the use case.** The tag format guidance, category list, searchability tips, and "discuss before proposing" guardrail create a good framework for productive AI-assisted lore management. The AI will produce entries that integrate well with the CJK bigram search system.

- **Proposal workflow is sound.** The accept/reject pattern with color-coded badges (green add, yellow edit, red delete), loading states, error recovery, and confirmation dialogs for destructive actions matches production-quality UX patterns.

- **Separation of "author mode" and "play mode" is correct.** The lore console lives at `/lore` as a separate page, not embedded in the game session. This prevents accidental lore edits during gameplay and gives the GM a focused workspace for world-building. The navigation link in the game header provides easy access without cluttering the play UI.

- **Integration with existing pipeline is clean.** Both the manual CRUD routes and the AI proposal `apply` route go through `_save_lore_entry()` / `delete_lore_entry()`, which update both `world_lore.json` and the SQLite search index. Changes made in the lore console will immediately affect the game's RAG-lite context injection on the next player turn.

- **Sub-group + toggle-all + alphabetical sort (commit 3) transforms usability at scale.** With 564 entries, default-all-collapsed with toggle-all is the right call. The three-level hierarchy (category > sub-group > entry) with distinct visual styling (gold > accent > default) creates a readable tree even at this scale.

### Verdict: PASS

The core feature is solid and addresses a real need -- managing 564 lore entries without a dedicated tool would be painful. The Round 2 changes addressed all my primary Round 1 concerns. The context window scaling issue (Medium) is worth tracking but does not block this PR since the current provider (Gemini) handles it fine. Ship it.

lainn9527 (commented):
## Addressing FE Review (Round 2) Feedback

**Retry button dead on proposal failure** — Fixed in 12a5384. Changed from direct `.btn-accept` event listener to event delegation on the `card` element using `ev.target.closest(".btn-accept")`. Dynamically-rendered retry buttons now work correctly.

All other FE review observations were acknowledged as non-blocking (no-debounce on search is fine for 41 entries, `box-sizing` is managed by the global reset, colon delimiter is intentional for CJK data).

PR #23, opened 2026-02-08
Collapse auto-play branches + cache-busting

lainn9527:
## Summary
- Auto-play branches now collapse into a single **「Auto-Play (N)」** group header in the drawer, collapsed by default
- Only expands if current branch is an auto-play branch
- Prevents auto-play branches from blocking main timeline branch list
- Adds cache-busting `?v=hash` to static file URLs (JS/CSS) so browser cache auto-invalidates on code changes

## Test plan
- [ ] Open drawer — auto-play branches should show under collapsible "Auto-Play (N)" header
- [ ] Header collapsed by default; click to expand/collapse
- [ ] Switch to an auto-play branch — section auto-expands
- [ ] Main timeline branches fully visible without scrolling past auto branches
- [ ] Hard-refresh no longer needed after code changes (cache-busting)

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -1146,13 +1146,15 @@ async function _btDeleteSelected() {
   const failures = [];
   for (const id of deduped) {
     try {
-      await API.deleteBranch(id);
+      const res = await API.deleteBranch(id);
+      if (!res.ok) failures.push(id);
     } catch (e) {
       failures.push(id);
     }
   }
   if (failures.length > 0) {
-    alert(`${failures.length} 個分支刪除失敗：${failures.join(", ")}`);
+    const names = failures.map(id => branches[id]?.name || id).join(", ");
+    alert(`${failures.length} 個分支刪除失敗：${names}`);
   }
   _btSelected.clear();
   await loadBranches();


lainn9527 (commented):
## Game Director Review — PR #23

### 1. Branch Tree Modal — Excellent Addition

The full tree visualization is a significant navigation upgrade. Players with complex branching narratives (dozens of branches deep) finally have a bird's-eye view. Specific positives:

- **Auto-scroll to current branch** (`scrollIntoView({ block: "center" })`) is the right call — players immediately orient themselves.
- **Single-child chain collapse** (inline rendering when `kids.length === 1`) avoids the "1000-deep linear chain" visual noise. Smart.
- **Cmd+B keyboard shortcut** — power users will appreciate this. Good discoverability through the 🌳 button for casual players.

### 2. Merge/Delete Safety — Mostly Good, One Concern

**What works:**
- Confirmation dialogs on every destructive action (single and batch).
- Descendant count warnings ("含 N 個子分支") — critical context given the user's history of losing 40 branches.
- `main` branch is protected (no checkbox, no action buttons).
- Confirm modal z-index (600) properly layers above the tree modal (500).

**Issue — Batch delete has no error handling per-item:**

In `_btDeleteSelected()`, the loop `for (const id of ids) { await API.deleteBranch(id); }` has no try/catch. If one delete fails mid-batch (e.g., deleting a child that was already cascade-deleted by its parent), the function silently continues or potentially throws, leaving the UI in a partial state. Additionally, if a user selects both a parent and its child, `countDescendants` will double-count, making the warning message misleadingly inflated.

**Recommendation:** Consider sorting selected IDs leaf-first (deepest descendants first) before deleting, and wrapping each call in a try/catch so partial failures don't break the batch. The descendant count in the confirmation could also deduplicate (exclude already-selected branches from the count).

### 3. Auto-Play Collapse — Clean Win

This directly addresses drawer clutter. The collapsible "Auto-Play (N)" section with:
- Collapsed by default
- Auto-expands only when viewing an auto-play branch
- Visual separator (border-top)

This is exactly the right UX. Auto-play branches are "background noise" for most players — they should be accessible but not dominant. LGTM on this feature.

### 4. Scroll-to-Branch-Point — Great for Story Navigation

When clicking a branch in the tree modal, scrolling to `branch_point_index + 1` (the first divergent message) is the narratively correct behavior. Players want to see "where did this story fork?" — not the end of the branch. The `scrollBlock: "start"` positioning puts the divergence point at the top of the viewport, which reads like "here's where this path begins."

The `requestAnimationFrame` + `setTimeout(50ms)` delay is a pragmatic fix for DOM rendering race conditions. Not elegant, but functional.

### 5. Overall Assessment

**No blocking issues.** This PR meaningfully improves the player experience for anyone managing more than a handful of branches. The tree modal transforms branch management from "scrolling through a flat list hoping you find the right one" to "visual navigation with full context."

**Minor items to consider (non-blocking):**

1. **Batch delete ordering** (described above) — edge case but worth hardening given the user's branch-loss history.
2. **Branch indicator change** — now shows raw `branch_id` instead of `branch.name`. This is a regression for readability; branch IDs like `branch_abc123de` are less meaningful than user-assigned names. The tooltip now has both, which is good, but the primary display got worse.
3. **No undo for delete** — given the 40-branch incident, a "soft delete with undo" or even a 5-second undo toast would be a stronger safety net than confirmation dialogs alone. (This is a larger feature request, not a PR blocker.)

LGTM with the batch-delete hardening as a suggested follow-up.

PR #24, opened 2026-02-08
Fix state sync: add current_phase + require GM STATE tag

lainn9527:
## Summary
- **Root cause**: `character_state.json`'s `current_status` drifts from narrative because `_extract_tags_async()` is probabilistic post-processing — it may or may not update `current_status` on scene transitions
- **Fix**: Require GM to output `<!--STATE {...} STATE-->` tag synchronously every response, with a new `current_phase` enum field for reliable scene positioning
- The existing regex pipeline (`_extract_state_tag`) extracts the tag **before** `state_snapshot` is captured (line 1209), so branch forks always inherit accurate state

### `app.py` (tracked)
- Add `current_phase` to `DEFAULT_CHARACTER_STATE` and `DEFAULT_CHARACTER_SCHEMA`
- Add `VALID_PHASES` constant with soft validation (log warning, never block)
- Backfill `current_phase = "主神空間"` in `_load_character_state()` for legacy branches
- Update `_extract_tags_async` prompt: add `current_phase` rules, remove `location` from forbidden list
- Add `log.warning` when GM response has no STATE tag (observability only)

### Data files (gitignored, applied directly to live `data/`)
- `system_prompt.txt`: Replace tag instruction — GM must output STATE tag with `current_phase` + `current_status` every response
- `character_schema.json`: Add `current_phase` field + `direct_overwrite_keys` entry
- `default_character_state.json`: Add `current_phase: "主神空間"`

## Data flow (before → after)
```
Before: GM narrative → (async LLM guesses current_status) → may or may not update
After:  GM narrative + STATE tag → regex sync extract → write state → snapshot captured → branch inherits correct state
```

## What doesn't change
- `_extract_state_tag()` regex — already supports `<!--STATE {...} STATE-->` format
- `_apply_state_update()` — already supports `direct_overwrite_keys`
- `skip_state=bool(state_updates)` — async extraction correctly falls back when no sync tag
- Frontend `renderCharacterStatus()` — schema-driven, auto-renders new field
- `auto_play.py` — reuses `_process_gm_response()`, benefits automatically

## Test plan
- [ ] Send message in Web UI, confirm GM response ends with STATE tag containing `current_phase`
- [ ] Trigger scene transition (e.g. enter dungeon), confirm `current_phase` updates to `副本中`
- [ ] Edit a user message to create branch, confirm new branch inherits correct `current_phase`
- [ ] Run `python auto_play.py --max-turns 5`, confirm STATE tags extracted and `current_phase` transitions
- [ ] Verify legacy branches without `current_phase` get backfilled on load

### DIFF
--- a/app.py
+++ b/app.py
@@ -656,7 +656,7 @@ def _extract_tags_async(story_id: str, branch_id: str, gm_text: str, msg_index: 
                 "- 文字型欄位直接覆蓋（如 `gene_lock: \"第二階\"`），值要簡短（5-20字）\n"
                 "- `current_phase` 只能是：主神空間/副本中/副本結算/傳送中/死亡\n"
                 "- 可以新增**永久性角色屬性**（如學會新體系時加 `修真境界`, `法力` 等）\n"
-                "- **禁止**新增臨時性/場景性欄位（如 threat_level, combat_status, escape_options 等一次性描述）\n"
+                "- **禁止**新增臨時性/場景性欄位（如 location, threat_level, combat_status, escape_options 等一次性描述）\n"
                 '- 角色死亡時 `current_phase` 設為 `"死亡"`，`current_status` 設為 `"end"`\n'
                 "格式：只填有變化的欄位。\n\n"
                 "## 輸出\n"
@@ -1184,7 +1184,7 @@ def _process_gm_response(gm_response: str, story_id: str, branch_id: str, msg_in
     for state_update in state_updates:
         _apply_state_update(story_id, branch_id, state_update)
     if not state_updates:
-        log.warning("GM response missing STATE tag (msg_index=%d)", msg_index)
+        log.info("GM response missing STATE tag (msg_index=%d)", msg_index)
 
     gm_response, lore_entries = _extract_lore_tag(gm_response)
     for lore_entry in lore_entries:


lainn9527 (commented):
**Backend Review**

Reviewed the diff (`main..HEAD`) covering the `current_phase` field addition, backfill logic, validation, prompt changes, and warning log.

---

**1. Backfill logic in `_load_character_state` -- Correct with one edge-case caveat**

The backfill at line 252 adds `current_phase = "主神空間"` when the key is missing. This runs before the existing `_save_json`, so the field gets persisted on first load. Correct for the common case.

Edge case to be aware of (non-blocking): if a legacy branch is currently mid-dungeon (the player is inside a dungeon instance), the backfill will incorrectly set the phase to `主神空間` instead of `副本中`. On the next GM response, the STATE tag should correct it, so this is self-healing. Acceptable trade-off -- just noting it for awareness.

**2. Thread safety of `VALID_PHASES` validation -- No issues**

`VALID_PHASES` is a module-level `frozenset`-like `set` literal that is never mutated after definition. The `in` membership test in `_apply_state_update` is a read-only operation, which is thread-safe in CPython. The validation is soft (log-only, never blocks), so even in the theoretical case of a race, the worst outcome is a missing or duplicated log line. No concerns here.

**3. `_extract_tags_async` prompt changes -- One concern**

- Adding `current_phase` enum constraint to the prompt: correct.
- Death handling (`current_phase = "死亡"` + `current_status = "end"`): correct, covers both fields.
- **Concern**: `location` was removed from the "forbidden temporary fields" example list. The rationale appears to be that `current_phase` replaces it, but `current_phase` is an enum of game phases (5 values), not a free-text location. With `location` no longer in the forbidden list, the async LLM extractor may start creating `location` fields again on character state (e.g., `location: "咒怨副本 — 伽椰子之家"`). This would re-introduce the state pollution that the forbidden list was designed to prevent. Consider keeping `location` in the forbidden examples.

**4. Warning log placement -- Needs adjustment**

The `log.warning("GM response missing STATE tag")` at line 1186-1187 fires immediately after `_extract_state_tag` regex extraction. However, the system has a two-stage extraction pipeline:
1. Sync regex extraction (`_extract_state_tag`) -- runs inline
2. Async LLM extraction (`_extract_tags_async`) -- runs in background thread, can also produce state updates (when `skip_state=False`)

The warning fires when stage 1 finds no STATE tag, but stage 2 (which was already the primary state extraction path before this PR) will still attempt state extraction in the background. This means the warning will fire on **every single GM response that doesn't contain a literal `<!--STATE ... STATE-->` tag**, which -- until the system prompt change fully takes effect -- will be the majority of responses. This could be extremely noisy.

Suggestion: either downgrade to `log.info` / `log.debug` during the transition period, or gate it behind a flag/counter so it only warns after repeated misses.

**5. Other observations**

- `direct_overwrite_keys` now includes `current_phase`: correct, this ensures `_apply_state_update_inner` handles it via the direct overwrite path at line 779-781.
- The `DEFAULT_CHARACTER_SCHEMA` field ordering places `current_phase` right after `name`: sensible for visibility in the frontend character panel.
- The invalid-phase validation in `_apply_state_update` uses `if phase and phase not in VALID_PHASES` -- the truthiness check means an empty string `""` would silently pass without logging. This is fine since an empty string won't overwrite anything meaningful, but worth noting.

**Summary**: Two actionable items:
1. Consider keeping `location` in the "forbidden temporary fields" example list in the `_extract_tags_async` prompt (line 659), since `current_phase` is not a 1:1 replacement for free-text location.
2. The STATE-tag-missing warning will be very noisy until all GM responses start including the tag. Consider downgrading severity or gating it.

Everything else looks correct.

lainn9527:
## Review responses

### Backend Review
Both actionable items addressed in 29bda25:
1. **`location` restored to forbidden list** — agreed, `current_phase` is a 5-value enum, not a free-text location replacement
2. **Log downgraded to `log.info`** — will be noisy during transition but useful for monitoring STATE tag adoption rate

### UI/UX Review
**"階段" vs "狀態" label confusion** — valid point. Will track as follow-up (label rename to "所在" or "場景階段"). For this PR, the semantic distinction is: 階段 = where (enum), 狀態 = what (free text).

### Game Director Review
- **Missing "副本準備" phase** — agreed that `主神空間` covers this transitional moment adequately
- **"Every response" is aggressive** — acceptable since async extraction is the fallback. The dual-path (sync tag + async extraction) provides resilience
- **`auto_play.py` phase desync** — good catch, tracked as follow-up. The regex-based phase detection in auto_play should eventually consume `current_phase` from character_state
- **Soft validation** — confirmed as intentional design choice

### Frontend Review
No changes needed — schema-driven rendering handles it correctly.

PR #25, opened 2026-02-08
UX fixes + lore source provenance

lainn9527:
## Summary
- **Loading indicator**: Show '主神系統處理中' with typing animation during regen and send streaming (previously showed blank)
- **Scroll fix**: Fix forced scroll-back after streaming completes — preserve user's current scroll position instead of jumping to saved position
- **Paragraph spacing**: Increase message paragraph margin from 0.8em to 1.2em for better readability
- **Lore source tracking**: Record where each lore entry was extracted from (branch, message index, timestamp, excerpt) and display in lore console
- **CLAUDE.md**: Add world timer, conversation compaction, and pending PR documentation

## Test plan
- [ ] Send a message → verify '主神系統處理中' typing dots appear while waiting
- [ ] Regen a GM response → verify same loading indicator
- [ ] During streaming, scroll down → verify page stays at scrolled position after streaming completes
- [ ] Check paragraph spacing in GM responses looks reasonable
- [ ] Open lore console → verify entries extracted by tag pipeline show source info
- [ ] Edit a lore entry → verify source provenance shown in modal

### DIFF
--- a/app.py
+++ b/app.py
@@ -2762,7 +2762,10 @@ def api_lore_entry_update():
                 if any(x.get("topic") == new_topic for x in lore):
                     return jsonify({"ok": False, "error": f"topic '{new_topic}' already exists"}), 409
                 delete_lore_entry(story_id, topic)
-            lore[i] = {"category": new_category, "topic": new_topic, "content": new_content}
+            updated = {"category": new_category, "topic": new_topic, "content": new_content}
+            if e.get("source"):
+                updated["source"] = e["source"]
+            lore[i] = updated
             _save_json(_story_lore_path(story_id), lore)
             upsert_lore_entry(story_id, lore[i])
             return jsonify({"ok": True, "entry": lore[i]})


lainn9527 (commented):
## Frontend Review — PR #25 UX Fixes + Lore Source Provenance

Reviewed all JS, CSS, and HTML changes. No issues found — LGTM.

### Loading Indicator (`app.js` lines 1506, 2395) — No XSS Risk

The `innerHTML` assignment uses a **hardcoded HTML string** with no user input interpolation:
```js
contentEl.innerHTML = '<span style="color:var(--text-dim);font-size:0.9rem">主神系統處理中<span class="typing-dots"><span></span><span></span><span></span></span></span>';
```
This is safe. The string is a static literal — no dynamic data flows into it. The `typing-dots` CSS animation (already defined in `style.css` lines 836-866) will render the bouncing dots correctly. Both usages (in `sendMessage` and `regenerateGmMessage`) are properly overwritten by the first `onChunk` callback which calls `markdownToHtml()`, so the indicator is cleanly replaced once streaming begins.

### Scroll Position Fix (`app.js` lines 1262-1270) — Correct

The change replaces `savedScrollTop` (captured at function entry, before DOM re-renders) with `currentScroll` (captured just before the `requestAnimationFrame`). This is the right fix for the `forcePreserve` path:

- `savedScrollTop` was captured at line 1205 at the very start of `switchToBranch()`, before async operations (API calls, DOM rebuilds) execute. By the time the `requestAnimationFrame` fires, the DOM has been fully rebuilt and the old `savedScrollTop` value is stale — it reflects the scroll position from before the branch switch.
- `currentScroll = container.scrollTop` on line 1264 captures the *current* position right before `rAF`, which is the user's actual scroll position after the DOM has been updated.
- Could `currentScroll` itself be "stale" by the time the `rAF` callback runs? In theory the user could scroll in the ~16ms between capture and rAF execution, but this is negligible and standard practice. The important thing is it reflects post-rebuild state, not pre-rebuild state.
- Note: the `preserveScroll` (non-force) path at lines 1273-1286 still uses `savedScrollTop` — this is correct for that path since it's used for variant switching where the intent is to restore the *original* position.

### Lore Source Display (`lore.js`) — Properly Escaped, No Injection Risk

All dynamic data is passed through `escapeHtml()`:

1. **`formatSourceLine()`** (lines 155-163): Builds a plain string from `source.branch_id`, `source.msg_index`, `source.timestamp`. The return value is always passed through `escapeHtml()` at the call sites.

2. **Entry preview** (line 176): `escapeHtml(formatSourceLine(e.source))` — double-safe since `formatSourceLine` returns a plain string and it's escaped before insertion.

3. **Modal source info** (lines 319-331): Both `formatSourceLine` output and `entry.source.excerpt` are wrapped in `escapeHtml()`. The excerpt comes from `gm_response[:100]` on the backend (line 1203 of `app.py`), which could contain arbitrary GM text, but escaping handles this correctly.

4. **Modal cleanup** (lines 308-309): `document.getElementById("modal-source-info").remove()` properly cleans up the previous source info block when reopening the modal, preventing DOM accumulation.

### CSS (`lore.css`) — No Conflicts

- `.lore-entry-source` (lines 280-285): New class scoped to the lore list. Uses existing CSS variables (`--text-dim`). No collision with `style.css`.
- `.modal-source-info`, `.modal-source-detail`, `.modal-source-excerpt` (lines 288-305): Scoped to the modal context. Use existing variables (`--input-bg`, `--border`, `--text-dim`, `--radius`). No naming conflicts found in `style.css`.

### Paragraph Spacing (`style.css`)

`.message .content p` margin change from `0.8em` to `1.2em` is a subtle readability improvement. The `p:first-child` override at the next line still applies.

### Backend Source Provenance

Both tag-pipeline extraction paths (direct `_process_gm_response` at line 1200 and async `_extract_tags_async` at line 699) attach the same `source` structure: `{branch_id, msg_index, excerpt, timestamp}`. The `_save_lore_entry` function preserves existing `source` if not provided in updates (lines 486-487), which is correct — manual edits via the API won't wipe provenance.

No issues found.

lainn9527 (commented):
## Game Director Review — PR #25

### Overall Verdict: LGTM (No Issues Found)

This PR makes four distinct improvements that collectively enhance the player experience. Each change is reviewed below from a narrative immersion and game-feel perspective.

---

### 1. Loading Indicator: "主神系統處理中" — Excellent

This is the standout change from a game design perspective. Previously, the GM message area went blank during streaming — a dead zone that breaks immersion and leaves the player uncertain whether anything is happening. The new loading text with animated typing dots solves this elegantly.

**Lore fit**: The phrase "主神系統處理中" (God System processing) is pitch-perfect. The system prompt (line 10) already establishes that 主神 system messages use 【】brackets with text like "【主神提示：任務更新……】". The loading indicator adopts this same in-world voice. The player sees the God System "thinking," not a generic spinner. This is the difference between a loading screen that says "Loading..." and one that says "The Animus is synchronizing..." — it keeps you inside the fiction.

The styling choice (dimmed text, 0.9rem, typing dots) is appropriately understated. It signals activity without demanding attention, leaving room for the streamed narrative to replace it naturally. Well done.

**One minor thought**: The indicator uses plain text rather than 【】brackets. This is actually the right call — using the brackets might make it look like an actual in-game system message, which could confuse players. The dim styling already differentiates it.

### 2. Scroll Behavior During Streaming — Correct Design Choice

The fix changes `forcePreserve` from restoring a *saved* scroll position to preserving the *current* scroll position. This matters because during streaming, the saved position becomes stale.

**For a narrative game, this is the right approach.** The existing `sendMessage()` flow already has smart scroll behavior: it auto-scrolls to new content only if the user was at the bottom AND hasn't scrolled away during streaming. The `forcePreserve` path (used during branch switches after streaming) should respect the same principle — if the player scrolled up to re-read earlier narrative during streaming, yanking them back down would be jarring.

In a novel reader, you would never force-scroll someone who is reading an earlier passage. The narrative will still be there when they scroll down. This respects the reader's agency.

### 3. Paragraph Spacing: 0.8em to 1.2em — Appropriate

For a text-heavy narrative RPG where GM responses can run 500+ words of scene description, dialogue, and action, paragraph spacing directly affects reading comfort.

- 0.8em was too tight for long narrative passages in Chinese, where paragraph breaks are the primary structural cue.
- 1.2em gives each paragraph room to breathe without fragmenting the narrative flow.
- On mobile, the override at `1em` with `text-indent: 2em` (line 1646-1648) creates a proper novel-reader feel. The desktop 1.2em complements this by giving similar visual separation through spacing rather than indentation.

For reference, most Chinese novel reader apps use 1.0-1.5em paragraph spacing. 1.2em sits comfortably in that range. This will make long dungeon scenes and combat sequences significantly more readable.

### 4. Lore Source Provenance — Useful, Not Noisy

This tracks where each lore entry was extracted from (branch, message index, timestamp, excerpt) and displays it in the lore console.

**Is this meta-game noise?** No. The lore console is already a meta-game tool — it is the GM's worldbuilding reference panel, not something shown during gameplay flow. Adding provenance here is like adding footnotes to a D&D sourcebook. The information lives behind a deliberate UI action (opening the lore console, clicking an entry).

**Why this matters for a branching narrative game:**
- With multiple branches and auto-play agents generating lore simultaneously, knowing *where* a lore entry came from prevents contradiction. If "副本世界觀：三體" was extracted from branch `auto_a3b2c1d4` at message #47, the GM (human or AI) can trace back to verify context.
- The `excerpt` field (first 100 chars of the GM response) gives just enough context to jog memory without requiring the player to navigate to the source message.
- The display is appropriately subtle: 0.75rem dimmed text with 0.7 opacity on the list, and a clean read-only block in the edit modal. It does not clutter the primary lore content.

**Implementation note**: The `_save_lore_entry()` change that preserves existing `source` on updates is correct — you don't want a lore content edit to erase the original provenance.

---

### Summary

| Change | Impact |
|--------|--------|
| 主神系統處理中 | Eliminates dead air, maintains fiction during load |
| Scroll preservation | Respects reader agency in a text-heavy game |
| Paragraph spacing | Better readability for long narrative passages |
| Lore provenance | Useful GM tool for branching narrative consistency |

No issues found. Ship it.

PR #26, opened 2026-02-08
Auto-cleanup failed branches + sibling delete button

lainn9527:
## Summary
- **Auto-cleanup failed branches**: Edit (✎) / Regen (↻) create a new branch before calling LLM. If the LLM call fails (e.g. Gemini key cooldown), the empty branch is now automatically deleted instead of lingering as a dead stub.
- **Sibling switcher ✕ button**: Hover over the `< 1/N >` switcher to reveal a ✕ button that deletes the current variant (with confirmation). Automatically switches to adjacent sibling after deletion.
- **Error callback branch reload**: Edit/regen error handlers now refresh the branch list so the drawer reflects cleanup immediately.

## Test plan
- [ ] Edit a message when LLM is down → verify no dead branch remains in branch tree
- [ ] Regen a GM response when LLM is down → same verification
- [ ] Hover over `< 1/2 >` switcher → ✕ appears
- [ ] Click ✕ → confirm → variant deleted, switches to adjacent sibling
- [ ] Verify ✕ does NOT appear on main branch variants
- [ ] Verify ✕ does NOT appear when only 1 variant exists (switcher hidden)
- [ ] Send a message when LLM is down → user message preserved (no cleanup, expected)

### DIFF
--- a/app.py
+++ b/app.py
@@ -1907,9 +1907,14 @@ def api_branches_edit():
     log.info("  context_search: %.0fms", (time.time() - t0) * 1000)
 
     t0 = time.time()
-    gm_response, _ = call_claude_gm(
-        augmented_edit, system_prompt, recent, session_id=None
-    )
+    try:
+        gm_response, _ = call_claude_gm(
+            augmented_edit, system_prompt, recent, session_id=None
+        )
+    except Exception as e:
+        log.info("/api/branches/edit EXCEPTION %s", e)
+        _cleanup_branch(story_id, branch_id)
+        return jsonify({"ok": False, "error": str(e)}), 500
     log.info("  claude_call: %.1fs", time.time() - t0)
 
     gm_response, image_info, snapshots = _process_gm_response(gm_response, story_id, branch_id, gm_msg_index)
@@ -2141,9 +2146,14 @@ def api_branches_regenerate():
     log.info("  context_search: %.0fms", (time.time() - t0) * 1000)
 
     t0 = time.time()
-    gm_response, _ = call_claude_gm(
-        augmented_regen, system_prompt, recent, session_id=None
-    )
+    try:
+        gm_response, _ = call_claude_gm(
+            augmented_regen, system_prompt, recent, session_id=None
+        )
+    except Exception as e:
+        log.info("/api/branches/regenerate EXCEPTION %s", e)
+        _cleanup_branch(story_id, branch_id)
+        return jsonify({"ok": False, "error": str(e)}), 500
     log.info("  claude_call: %.1fs", time.time() - t0)
 
     gm_msg_index = branch_point_index + 1
--- a/static/app.js
+++ b/static/app.js
@@ -1772,19 +1772,28 @@ function renderMessages(messages) {
         delBtn.title = "刪除此版本";
         delBtn.addEventListener("click", async (e) => {
           e.stopPropagation();
-          if (!(await showConfirm("刪除此版本？"))) return;
           const branchToDelete = currentVariant.branch_id;
-          const res = await API.deleteBranch(branchToDelete);
-          if (res.ok) {
-            const adjIdx = group.current_variant >= 2 ? group.current_variant - 2 : 0;
-            const adjacent = group.variants[adjIdx];
-            await loadBranches();
-            if (adjacent && adjacent.branch_id !== branchToDelete) {
-              await switchToBranch(adjacent.branch_id, { preserveScroll: true });
+          const descCount = countDescendants(branchToDelete);
+          let msg = "刪除此版本？";
+          if (descCount > 0) msg += `\n（含 ${descCount} 個子分支也會一併刪除）`;
+          if (!(await showConfirm(msg))) return;
+          try {
+            const res = await API.deleteBranch(branchToDelete);
+            if (res.ok) {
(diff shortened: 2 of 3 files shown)


lainn9527 (commented):
## Frontend Review — Sibling Delete Button + Error Callbacks

### 1. `adjIdx` Calculation: Can It Go Negative? (No Issue)

```js
const adjIdx = group.current_variant >= 2 ? group.current_variant - 2 : 0;
```

This is safe. The ternary guard `current_variant >= 2` ensures `current_variant - 2 >= 0`. The else branch yields `0`. Tracing all cases:

- `current_variant = 1` (first variant): `adjIdx = 0`. `variants[0]` is the deleted branch itself, so `adjacent.branch_id !== branchToDelete` fails, falls through to `switchToBranch("main")`. Correct.
- `current_variant = 2`: `adjIdx = 0`. Picks the previous variant. Correct.
- `current_variant = N` (N >= 2): `adjIdx = N - 2`. Always picks the preceding variant. Correct.
- `current_variant = 0` (edge case -- no variant marked as current in backend): `group.variants[-1]` is `undefined` in JS, so `currentVariant` is `undefined`, the outer `if (currentVariant && ...)` is falsy, and the delete button is never rendered. Safe.

No negative index issue.

### 2. Missing Error Handling on `API.deleteBranch()` Failure (Actionable)

**Lines 1777-1788:** The `if (res.ok)` path is handled, but there is no `else` branch. Compare with the existing patterns at lines 738-747 and 1037-1047:

```js
// Existing pattern (line 745-747):
} else {
  alert(res.error || "刪除失敗");
}
```

The new delete button silently does nothing on failure. Since this is a user-initiated destructive action (with a confirmation dialog!), silent failure is confusing -- the user confirmed deletion, nothing visibly happened, and there is no feedback.

**Additionally**, there is no `try-catch` around the `await API.deleteBranch()` call. `API.deleteBranch` is `fetch(...).then(r => r.json())`, so a network error will reject the promise and the unhandled rejection will be swallowed silently.

**Suggestion:** Add both an `else` branch and a `try-catch`, consistent with the existing codebase pattern:

```js
delBtn.addEventListener("click", async (e) => {
  e.stopPropagation();
  if (!(await showConfirm("刪除此版本？"))) return;
  const branchToDelete = currentVariant.branch_id;
  try {
    const res = await API.deleteBranch(branchToDelete);
    if (res.ok) {
      const adjIdx = group.current_variant >= 2 ? group.current_variant - 2 : 0;
      const adjacent = group.variants[adjIdx];
      await loadBranches();
      if (adjacent && adjacent.branch_id !== branchToDelete) {
        await switchToBranch(adjacent.branch_id, { preserveScroll: true });
      } else {
        await switchToBranch("main");
      }
      renderBranchList();
    } else {
      alert(res.error || "刪除失敗");
    }
  } catch (err) {
    alert("刪除失敗：" + err.message);
  }
});
```

### 3. `loadBranches().then(() => renderBranchList())` in Edit/Regen Error Handlers (No Issue)

**Lines 1479, 1554:** The backend `_cleanup_branch()` already deleted the failed branch server-side, so calling `loadBranches()` to re-fetch the branch tree is correct -- the drawer needs to reflect that the dead branch is gone. The `.then(() => renderBranchList())` ensures the drawer re-renders after the fresh data arrives. The `alert()` on the next line fires immediately (before the promise resolves), which means the alert pops up while `loadBranches` is in-flight. This is fine from a UX standpoint -- the alert is dismissed by the user, and by that time `loadBranches` has almost certainly completed.

No race condition concern here.

### 4. Sequence in Delete Button: `loadBranches` -> `switchToBranch` -> `renderBranchList` (No Issue)

```js
await loadBranches();                          // 1. Refresh branch data
await switchToBranch(adjacent.branch_id, ...); // 2. Switch view (loads messages, re-renders)
renderBranchList();                            // 3. Re-render drawer
```

All three are `await`ed in sequence (no `.then()` parallelism), so there's no race condition. `switchToBranch` internally calls `loadMessages()` which calls `renderMessages()`, which will read the updated `siblingGroups` from the new branch's perspective. `renderBranchList()` at the end updates the drawer. This is the same pattern used in the existing delete handlers (lines 740-744, 1039-1044). Correct.

### 5. `showConfirm` Exists (Verified)

`showConfirm` is defined at line 5 of `app.js` as a Promise-based modal dialog. It's used in 7+ other places in the codebase. No issue.

### 6. Stale Closure on `currentVariant` (No Issue, but Worth Noting)

The `currentVariant` and `group` objects are captured in the click handler closure at render time. If the user somehow triggers a re-render between hovering and clicking (e.g., a background NPC activity poll updates the UI), the closure would hold stale data. However, this is a theoretical concern -- the same pattern exists for the left/right arrow buttons in the switcher (lines 1738-1743, 1754-1758), and it works fine in practice because `renderMessages` replaces the entire DOM.

### Summary

| # | Area | Severity | Status |
|---|------|----------|--------|
| 1 | `adjIdx` negative index | -- | Safe, verified all cases |
| 2 | Missing `else` + `try-catch` on delete failure | Medium | **Actionable** -- silent failure after user confirmation is bad UX |
| 3 | Error handler `loadBranches` placement | -- | Correct |
| 4 | Delete sequence race conditions | -- | No issue, properly sequenced |
| 5 | `showConfirm` existence | -- | Verified (line 5) |

PR #28, opened 2026-02-08
Lore cleanup: dedup, split long entries, fix extraction prompt

lainn9527:
## Summary
- **`scripts/lore_cleanup.py`**: One-time cleanup script for `world_lore.json` that ran and reduced 576 bloated entries down to focused sub-entries
  - Phase 1: Exact dedup by topic (removed 0 — topics were all unique but fragmented)
  - Phase 2: Absorbed 212 micro-fragments (<300 chars) into their parent entries by topic-prefix matching
  - Phase 3: Remapped 29 entries from non-canonical categories (道具→商城, 基本屬性→體系, etc.)
  - Phase 4: Split 27 long entries (2000-15000 chars) into focused 200-800 char sub-entries via LLM calls
  - Phase 5: Final dedup, validation, save, and SQLite FTS index rebuild
- **`app.py`**: Fix `_extract_tags_async()` to prevent recurrence
  - Improved extraction prompt: guides LLM to upsert existing entries instead of creating new ones, enforces 200-800 char limit
  - Removed topic-existence guard so `_save_lore_entry()` can perform upserts (it already handles dedup by topic)

**Before**: 576 entries, 23 entries >2000 chars (up to 15623 chars), search results truncated at 800 chars  
**After**: 758 focused entries, 0 entries >2000 chars, search returns complete focused sub-entries

## Test plan
- [x] Ran `--dry-run` first to verify analysis
- [x] Ran `--skip-split` to verify dedup/remap without LLM calls
- [x] Ran full cleanup — all 27 long entries split successfully (1 timeout on 鬥氣 chunk 2, rest succeeded)
- [x] Verified search: `search_lore('story_original', '基因鎖開啟')` returns focused 166-297 char entries instead of truncated 6347-char parent
- [x] Verified no duplicate topics in final output
- [ ] Play a turn in web UI — verify `[相關世界設定]` section is non-truncated and relevant

### DIFF
--- a/scripts/lore_cleanup.py
+++ b/scripts/lore_cleanup.py
@@ -279,7 +279,7 @@ def _parse_json_array(text: str) -> list[dict]:
     except json.JSONDecodeError:
         pass
 
-    # Fallback: find first JSON array
+    # Fallback: find JSON array (greedy — json.loads validates correctness)
     m = re.search(r"\[[\s\S]*\]", text)
     if m:
         try:
@@ -335,10 +335,17 @@ def split_long_entries(lore: list[dict], delay: float = 3.0, dry_run: bool = Fal
             result = call_oneshot(prompt)
             if result:
                 parsed = _parse_json_array(result)
-                sub_entries.extend(parsed)
-                log.info("    Chunk %d → %d sub-entries", ci, len(parsed))
+                if parsed:
+                    sub_entries.extend(parsed)
+                    log.info("    Chunk %d → %d sub-entries", ci, len(parsed))
+                else:
+                    log.warning("    Chunk %d: LLM returned non-parseable response, aborting split", ci)
+                    sub_entries = []  # discard partial results to avoid data loss
+                    break
             else:
-                log.warning("    Chunk %d: empty LLM response", ci)
+                log.warning("    Chunk %d: empty LLM response, aborting split", ci)
+                sub_entries = []  # discard partial results to avoid data loss
+                break
             if ci < len(chunks):
                 time.sleep(delay)
 


lainn9527 (commented):
**Frontend Review**

Reviewed all frontend-facing changes for compatibility. **No frontend issues found.**

**Lore data format -- no breaking changes:**
- `lore.js` consumes only `category`, `topic`, and `content` fields from `/api/lore/all` responses (lines 62, 155-174). It never reads a `source` field.
- The removal of `source` provenance tracking from `_save_lore_entry()`, `_extract_tags_async()`, `_process_gm_response()`, and `api_lore_entry_update()` is purely a backend-internal change. The `source` field was stored in `world_lore.json` but was never surfaced to any API response consumed by the frontend. No JS/CSS/HTML code references `source` on lore objects.

**Search output format -- unchanged:**
- `search_relevant_lore()` in `lore_db.py` returns formatted text injected into augmented messages (backend-to-GM context). This is never exposed to the frontend. The function's output format (lines 283-303 of `lore_db.py`) is unchanged.
- `search_lore()` returns `{category, topic, content, tags, score}` -- same as before. No fields added or removed.

**`_cleanup_branch` removal -- acceptable for frontend:**
- The removed `_cleanup_branch()` was called on edit/regen error paths. Previously, on LLM failure, the backend would delete the newly created branch and return a 500 error. Now the branch persists on failure but the error response shape (`{"ok": false, "error": "..."}`) remains the same.
- For SSE streaming paths, the `{"type": "error", "message": "..."}` event format is unchanged (line 222 of `app.js` handles this).
- The only behavioral difference: orphan branches may accumulate on repeated failures. This is a backend-side concern, not a frontend one. The frontend will still receive the same error responses and handle them correctly.

**Lore cleanup script (`scripts/lore_cleanup.py`):**
- One-time offline script, no API changes. Output format preserves the standard `{category, topic, content}` shape. The sub-group rendering in `lore.js` (lines 108-153) already handles topic prefixes like "parent:sub" via `parseTopicPrefix()`, so the new split entries with topics like "基因鎖：概述" will actually render *better* in the lore console's sub-group accordion UI.

**Verdict: No frontend issues.** The lore data shape is stable, and the lore console UI will naturally benefit from the cleaner, shorter entries.

lainn9527 (commented):
**Backend Review**

Overall: solid one-time cleanup script with good phasing, dry-run support, and backup. The app.py changes to enable upsert and improve the extraction prompt are well-motivated. A few issues to flag:

---

**Critical**

1. **No retry on LLM failure in `split_long_entries`** (lore_cleanup.py L330-341)

   `call_oneshot()` returns an empty string on failure (timeout, API error, etc.). The script logs a warning but moves on, meaning that chunk's content is silently lost. Given the PR description mentions a timeout on the 鬥氣 chunk, this has already been observed in practice.

   The original entry is kept only if *all* chunks produce nothing (`sub_entries` stays empty). But if chunk 1 succeeds and chunk 2 times out, the original is replaced with a partial result, losing the chunk 2 content permanently.

   Suggestion: If any chunk returns empty, abort the split for that entry and keep the original. Or add a simple retry (1-2 attempts with backoff).

   ```python
   # After line 341
   if not result or not _parse_json_array(result):
       log.warning("    Chunk %d failed, aborting split for '%s'", ci, topic)
       sub_entries = []  # discard partial results
       break
   ```

2. **Removing try/except from non-streaming edit/regen routes** (app.py L1897-1900, L2130-2133)

   The old code caught `call_claude_gm` exceptions and returned a JSON error response. Now an uncaught exception will produce a bare Flask 500 (HTML error page, not JSON). The main `api_send` route also lacks this try/except, so this is at least consistent -- but it means all three routes now give the frontend an HTML error it can't parse. This is arguably a pre-existing issue, not a regression, but worth noting since the old behavior was strictly better for edit/regen.

---

**Suggestions**

3. **Fragment absorption can create topic collisions** (lore_cleanup.py L180-188)

   When a fragment is absorbed, its content is appended to the parent, but the parent entry's `content` may now exceed 2000 chars, making it a candidate for Phase 4 splitting. This works correctly because Phase 4 runs after Phase 2 -- good sequencing. However, there is no check that two different fragments mapping to the same parent could cause issues. This seems fine because `topic_to_idx` maps to a single index, but worth a comment noting the intentional ordering dependency.

4. **`_find_parent_topic` could match false positives** (lore_cleanup.py L137-140)

   The `topic.startswith(candidate)` check is purely prefix-based. For example, a topic "精神力量的覺醒" would match parent "精神力" even if they are conceptually unrelated. Given the 300-char threshold, this is unlikely to cause data loss with real lore entries, but consider adding a minimum parent topic length (e.g., >= 2 chars) to avoid very short false matches.

5. **`_parse_json_array` greedy regex** (lore_cleanup.py L283)

   `re.search(r"\[[\s\S]*\]", text)` is greedy and will match from the first `[` to the last `]` in the entire text. If the LLM outputs explanatory text after the JSON array containing `]`, this could capture garbage. Consider using a non-greedy match or `re.search(r"\[[\s\S]*?\](?=\s*$)", text)`.

6. **Backup file naming relies on `.json` suffix** (lore_cleanup.py L90)

   `src.replace(".json", f".backup.{ts}.json")` -- `str.replace` replaces ALL occurrences. If the path somehow contains `.json` earlier (e.g., `data/stories/story.json.bak/world_lore.json`), this would produce a malformed path. Minor, but `src.rsplit(".json", 1)` or `Path(src).with_suffix(...)` would be safer.

---

**Nits**

7. The `source` provenance field is fully removed from lore entries (app.py diff). This loses the audit trail of which branch/message produced each lore entry. If this is intentional (entries are now upserted so the original source is overwritten anyway), a brief comment in the commit message would help future readers understand why.

8. `_cleanup_branch` is removed from app.py. Confirmed: no remaining callers exist. Clean removal.

9. lore_cleanup.py L427 log says "Rebuilt SQLite FTS index" but `lore_db.py`'s `rebuild_index` uses a `lore` table + FTS5 triggers, not pure FTS rebuild. Minor log accuracy issue.

lainn9527:
**Re: Game Director — Unrelated Changes Bundled In**

The items flagged (1-8) are not part of this PR — they are changes on `main` that show up in the diff because `gh pr` compares branch vs base. This PR contains exactly 2 commits:

1. `df181f8` — Add lore cleanup script + fix extraction prompt (new `scripts/lore_cleanup.py`, 2 edits in `app.py`)
2. `0edb82e` — Fix abort-on-chunk-failure (1 edit in `scripts/lore_cleanup.py`)

The `_cleanup_branch` removal, Gemini max_tokens change, sibling delete button, loading indicator changes etc. are all already on `main` from prior PRs.

PR #29, opened 2026-02-08
Reparent children on branch delete

lainn9527:
## Summary
- When deleting a branch with children, reparent them to the grandparent instead of cascade deleting
- Materialize inherited messages into child's delta when needed (Case A: `bp(child) >= bp(deleted)`)
- Skip message changes when child doesn't depend on deleted branch's messages (Case B/C)
- Update confirm dialog text: "子分支將重新掛載" instead of "一併刪除"

## Motivation
Edit (✎) creates a **child** branch of the original. Users want to delete the old version while keeping the edited version — currently impossible because cascade delete removes children too.

## Cases
| Case | Condition | Action |
|------|-----------|--------|
| A | `bp(child) >= bp(deleted)` | Materialize inherited msgs into child delta, set `child.bp = deleted.bp` |
| B | `bp(child) < bp(deleted)` | Just reparent, no message changes |
| C | Blank child (`bp=-1`) | Just reparent, no message changes |

## Test plan
- [ ] Create A → B → C chain. Delete B → verify C's `get_full_timeline` is unchanged
- [ ] A → B → {C1(bp=70), C2(bp=75)}. Delete B → verify C1 and C2 deltas are correct
- [ ] Delete branch with children → branch tree and sibling switcher display correctly
- [ ] Delete leaf branch → behavior unchanged
- [ ] Blank child reparent → no inherited messages, parent updated correctly

### DIFF
--- a/app.py
+++ b/app.py
@@ -2483,19 +2483,23 @@ def api_branches_delete(branch_id):
     deleted_parent = branch.get("parent_branch_id", "main") or "main"
     deleted_bp = branch.get("branch_point_index")
 
-    # Find direct non-deleted, non-merged children
-    children = [
+    # Reparent ALL children (including deleted/merged) to prevent dangling refs
+    all_children = [
         b for b in branches.values()
         if b.get("parent_branch_id") == branch_id
-        and not b.get("deleted")
-        and not b.get("merged")
     ]
 
-    if children:
+    # Only materialize messages for active (non-deleted, non-merged) children
+    active_children = [
+        b for b in all_children
+        if not b.get("deleted") and not b.get("merged")
+    ]
+
+    if active_children:
         # Load deleted branch's delta messages before we remove the branch dir
         deleted_delta = _load_json(_story_messages_path(story_id, branch_id), [])
 
-        for child in children:
+        for child in active_children:
             child_id = child["id"]
             child_bp = child.get("branch_point_index")
 
@@ -2515,8 +2519,9 @@ def api_branches_delete(branch_id):
                     pass
             # else: blank child (bp=-1) — no messages inherited, bp stays -1
 
-            # Always reparent to grandparent
-            child["parent_branch_id"] = deleted_parent
+    # Reparent all children (including deleted/merged) to grandparent
+    for child in all_children:
+        child["parent_branch_id"] = deleted_parent
 
     # Delete the branch itself (preserve existing was_main soft-delete logic)
     if branch.get("was_main"):
@@ -2534,7 +2539,7 @@ def api_branches_delete(branch_id):
 
     _save_tree(story_id, tree)
 
-    return jsonify({"ok": True})
+    return jsonify({"ok": True, "switch_to": deleted_parent})
 
 
 # ---------------------------------------------------------------------------
--- a/static/app.js
+++ b/static/app.js
@@ -757,7 +757,7 @@ function createBranchItem(branch, depth, hasChildren, isExpanded) {
       if (res.ok) {
         await loadBranches();
         if (currentBranchId === branch.id || !branches[currentBranchId]) {
-          await switchToBranch("main");
+          await switchToBranch(res.switch_to || "main");
         }
         renderBranchList();
       } else {
@@ -1050,13 +1050,13 @@ function _btRenderTree(container, modal) {
         e.stopPropagation();
         const desc = countDescendants(branch.id);
         let msg = `刪除「${branch.name}」？`;
-        if (desc > 0) msg += `\n（${desc} 個子分支將重新掛載）`;
+        if (desc > 0) msg += `\n（${desc} 個子分支將重新掛載到上層）`;
         if (!(await showConfirm(msg))) return;
         const res = await API.deleteBranch(branch.id);
         if (res.ok) {
           await loadBranches();
           if (currentBranchId === branch.id || !branches[currentBranchId]) {
-            await switchToBranch("main");
+            await switchToBranch(res.switch_to || "main");
           }
           _btRenderTree(container, modal);
           renderBranchList();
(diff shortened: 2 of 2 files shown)


lainn9527 (commented):
**UI/UX Review**

Reviewed the 4 confirm dialog changes, post-delete navigation, drawer behavior, sibling switcher implications, and batch delete flow.

---

### 1. Confirm Dialog Wording

The 4 dialogs changed from cascade-delete language to reparent language:

| Location | Old | New |
|----------|-----|-----|
| Drawer delete (line 753) | `N 個子分支也會一併刪除` | `N 個子分支將重新掛載到上層` |
| Branch tree modal (line 1053) | `含 N 個子分支` | `N 個子分支將重新掛載` |
| Batch delete (line 1158) | `另含 N 個子分支也會一併刪除` | `另含 N 個子分支將重新掛載` |
| Sibling switcher (line 1796) | `含 N 個子分支也會一併刪除` | `N 個子分支將重新掛載` |

**Issue: Inconsistent phrasing across the 4 locations.** The drawer delete says `重新掛載到上層` (reparent to parent), which is the clearest. The other three just say `重新掛載` without specifying *where*. For a potentially confusing operation like reparenting, all 4 dialogs should use the same full phrasing `將重新掛載到上層` so users understand what will happen. The term `重新掛載` alone is somewhat technical — consider also saying `（子分支不會被刪除，將移至上層分支下）` for extra clarity, but at minimum the 4 dialogs should be consistent.

**Severity: Minor.** Functional but could confuse users who see different messages in different contexts.

---

### 2. Post-Delete Active Branch: Backend vs Frontend Mismatch

**Issue (Major): The backend sets `active_branch_id = deleted_parent`, but the frontend ignores this and independently switches to `"main"` in 3 of 4 locations.**

Backend (app.py line 2532-2533):
```python
if tree.get("active_branch_id") == branch_id:
    tree["active_branch_id"] = deleted_parent
```

Frontend behavior after successful delete:
- **Drawer** (line 759-760): `switchToBranch("main")` — always goes to main
- **Branch tree modal** (line 1058-1059): `switchToBranch("main")` — always goes to main
- **Batch delete** (line 1179): `switchToBranch("main")` — always goes to main
- **Sibling switcher** (line 1801-1808): Tries adjacent sibling first, then falls back to `"main"` — never uses parent

The backend's intent to navigate to `deleted_parent` is completely overridden by the frontend. This means:

1. If you delete a deeply nested branch while viewing it, you jump all the way to `main` instead of the natural parent. This is jarring — the user probably wants to continue in the context of the parent branch.
2. The backend response `{"ok": True}` does not return `deleted_parent`, so even if the frontend wanted to use it, the information is not available.

**Recommendation:** Return `deleted_parent` in the API response:
```python
return jsonify({"ok": True, "switch_to": deleted_parent})
```
Then update all 4 frontend locations to use `res.switch_to` (or fall back to `"main"`):
```js
await switchToBranch(res.switch_to || "main");
```

This makes the navigation feel natural — after deleting a branch, the user lands on the parent where the reparented children now live, which reinforces the mental model that children were "moved up" rather than lost.

---

### 3. Drawer Behavior After Reparent

After deletion and `loadBranches()` + `renderBranchList()`, reparented children appear under the grandparent. This is correct behavior and should not cause confusion *if* the user understood the confirm dialog. No issues here — the drawer re-renders from the updated tree data, so the visual state is always consistent.

One minor note: if the user deletes a branch while viewing a different branch, the drawer silently reorganizes. This is fine since the user initiated the action.

---

### 4. Sibling Switcher After Reparent

**Potential confusion (Minor):** When a child is reparented from branch B to grandparent A, and the child's `branch_point_index` is adjusted to `deleted_bp` (Case A in backend), the child may now share the same `(parent_id, branch_point_index)` as other existing branches under grandparent A. This means the reparented child could appear in a sibling group `< 1/N >` switcher alongside unrelated branches that happened to fork at the same index on the grandparent.

The `_get_sibling_groups` logic (line 940-951) groups by `(parent_id, bp_index)`, so this grouping is technically correct from a data perspective. However, from a user's perspective, a branch that was originally a "grandchild" now appearing as a sibling of its former "uncle" branches could be surprising.

**Mitigation:** This is an edge case that only occurs when (a) the deleted branch had children that forked within its delta range AND (b) the grandparent already had other children at the same fork index. In practice this is rare. No code change needed, but worth noting in case users report confusion.

---

### 5. Batch Delete Ordering

The batch delete sorts `leaf-first` (line 1162):
```js
deduped.sort((a, b) => countDescendants(a) - countDescendants(b));
```

With reparent-on-delete, this sorting actually becomes important in a new way. Consider selecting branches A and B where B is a child of A:
- The `deduped` filter (lines 1146-1153) removes B since its ancestor A is also selected — so only A is deleted.
- A's children (including B) get reparented to A's parent.
- But B was selected for deletion! It now exists under A's parent but was not actually deleted.

Wait — re-reading the dedup logic: if B's ancestor A is in the selected set, B is *filtered out* entirely. So B is never deleted. Under the old cascade behavior, B would have been deleted as part of A's cascade. Under the new reparent behavior, **B survives** because:
1. Dedup removes it from the delete list (ancestor A is selected)
2. A's delete reparents B to A's parent
3. B is now under A's parent, alive and well

**This is actually correct behavior for reparent-on-delete** — the dedup assumes children are handled by the parent's delete. Under cascade, "handled" meant "also deleted." Under reparent, "handled" means "reparented." The user selected both A and B, but B only survives because the system reparents it. The user might expect B to also be deleted.

**Issue (Minor):** The confirm dialog says `確定刪除 N 個分支？` showing the deduped count, plus `另含 M 個子分支將重新掛載`. If the user selected both A and B, the dialog shows "delete 1 branch" (just A) and "1 child branch will be reparented" (B). The user might be confused that B, which they explicitly selected, is being reparented instead of deleted. Consider adding a note like `（已選取的子分支會隨上層一起處理）` to the dialog when dedup reduces the count.

---

### Summary

| Issue | Severity | Action Needed |
|-------|----------|---------------|
| Inconsistent confirm dialog wording across 4 locations | Minor | Unify to `將重新掛載到上層` |
| Frontend always switches to `main` instead of `deleted_parent` | **Major** | Return `switch_to` in API response; use it in frontend |
| Batch delete: explicitly selected children survive via reparent | Minor | Consider clarifying in dialog, or accept as intended behavior |

**Requesting changes** for issue #2 — the backend/frontend mismatch on post-delete navigation is a meaningful UX regression. The other items are minor polish.

PR #158, opened 2026-03-07
Refactor app.py into modular helpers

lainn9527:
## Summary
- extract shared file I/O, path helpers, registry/tree helpers, branch message locks, and pending extract tracking into `story_io.py`
- keep `app.py` re-export compatibility for existing imports, including `auto_play.py`
- add a shared test helper to patch path constants across extracted modules and align compaction tests/docs with the intentional `10/10` thresholds

## Testing
- python3 -m pytest tests/test_compaction.py -q
- python3 -m pytest tests/ -x

PR #159, opened 2026-03-07
Refactor app_helpers into focused modules

lainn9527:
## Summary
- extract gm plan helpers into `gm_plan.py`
- extract legacy migration/startup helpers into `migrations.py`
- extract debug panel helpers into `debug_helpers.py`
- extract state update/event normalization helpers into `state_updates.py` and wire `gm_pipeline` to the new state constants

## Testing
- `python3 -m pytest tests/test_extract_tags_async.py tests/test_api_routes.py -x`
- `python3 -m pytest tests/test_api_routes.py -x`
- `python3 -m pytest tests/test_debug_api.py -x`
- `python3 -m pytest tests/test_state_update.py tests/test_state_review.py tests/test_extract_tags_async.py -x`
- `python3 -m pytest tests/test_debug_api.py -x`
- `python3 -m pytest tests/ -x`
- `python3 -c "from auto_play import *"`


lainn9527:
# PR #159 Review: Refactor app_helpers into focused modules

## Summary

This PR continues the modularization effort from PR #158 by extracting four focused modules from the monolithic `app_helpers.py`:

| New Module | Responsibility | Lines |
|---|---|---|
| `gm_plan.py` | GM plan CRUD, payoff tracking, prompt injection | 334 |
| `migrations.py` | Legacy data migrations & startup init | 370 |
| `debug_helpers.py` | Debug panel chat, directives, action apply | 392 |
| `state_updates.py` | State update pipeline, event ops, map/list normalization | 613 |

Additionally, `gm_pipeline.py` is updated to directly import constants (`_SCENE_KEYS`, `_INSTRUCTION_KEYS`, `_EVENT_STATUS_ORDER`, `_is_numeric_value`) from `state_updates.py` instead of using `app_module.*`.

---

## ✅ Positives

1. **Clean module boundaries** — Each module has a clear single responsibility and explicit `__all__` exports. Good separation.
2. **gm_pipeline cleanup** — Replacing `app_module._SCENE_KEYS` with direct imports from `state_updates` is a definite improvement; removes unnecessary late-binding lookups.
3. **Tests updated** — `conftest.py` properly patches the new `migrations` module for path overrides.
4. **Behavior-preserving** — The code appears to be a pure mechanical move with minimal logic changes.

---

## 🔍 Issues & Suggestions

### 1. 🔴 `_parse_env_int` duplicated 3 times

`_parse_env_int` is defined identically in:
- [app_helpers.py:L195](file:///Users/eddylai/story-refactor-app-helpers/app_helpers.py#L195)
- [gm_plan.py:L13](file:///Users/eddylai/story-refactor-app-helpers/gm_plan.py#L13)
- [debug_helpers.py:L31](file:///Users/eddylai/story-refactor-app-helpers/debug_helpers.py#L31)

This is a DRY violation. Consider placing it in a shared utils module (e.g. `story_io.py` or a new `config_utils.py`) and importing it. If this is intentional to avoid cross-module dependencies, at least leave a comment explaining the rationale.

### 2. 🟡 `_DynamicPath` — clever but potentially fragile

```python
class _DynamicPath(os.PathLike):
    def __init__(self, resolver):
        self._resolver = resolver
```

`_DynamicPath` in `migrations.py` is an interesting solution for lazy path resolution after test monkeypatching. However:

- It lacks `__eq__` and `__hash__`, so comparing paths (e.g. `if LEGACY_PARSED_PATH == some_path:`) may silently fail.
- It may surprise callers who expect a plain `str`. For example, `os.path.exists()` works (via `__fspath__`), but `str.startswith()` won't unless `__str__` is called explicitly.

> [!TIP]
> Consider adding a brief docstring explaining why `_DynamicPath` is needed (test patching of `story_io.DATA_DIR` after module import).

### 3. 🟡 Circular import in `state_updates.py` (deferred)

```python
# state_updates.py:L573
def _apply_state_update(story_id: str, branch_id: str, update: dict):
    ...
    from gm_pipeline import _normalize_state_async, _run_state_gate
```

The deferred `from gm_pipeline import ...` inside `_apply_state_update` creates a circular dependency: `gm_pipeline → state_updates` (top-level) and `state_updates → gm_pipeline` (deferred). This works at runtime, but:

- It makes the dependency graph harder to reason about.
- It will break if anyone tries to add a top-level import in the future.

This is an existing pattern from `app_helpers`, so not a regression — but worth noting for future cleanup.

### 4. 🟡 `migrations.py` still depends on `app_helpers` at runtime

```python
# migrations.py:L145
from app_helpers import DEFAULT_CHARACTER_SCHEMA, DEFAULT_CHARACTER_STATE
```

This deferred import inside `_migrate_to_stories()` creates a circular dependency path: `app_helpers → migrations` (via `from migrations import *`) and `migrations → app_helpers` (deferred). Like #3, it works but is fragile. Consider moving `DEFAULT_CHARACTER_SCHEMA`/`DEFAULT_CHARACTER_STATE` to `character_state.py` or `story_io.py` to break the cycle.

### 5. 🟢 Minor: `save_parsed` call signature changed in `core_routes.py`

```python
# Before (in base branch)
app_module.save_parsed()

# After
app_module.save_parsed(
    app_module.parse_conversation(os.fspath(app_module.CONVERSATION_PATH)),
    output=os.fspath(app_module.LEGACY_PARSED_PATH),
)
```

This changes the calling convention — the old `save_parsed()` was a no-arg call that internally resolved paths. The new version requires explicit `os.fspath()` wrapping because `CONVERSATION_PATH` is now a `_DynamicPath` instead of a plain `str`. This is correct but is a subtle behavioral change that could affect other callers of `save_parsed()`. Worth verifying no other call sites exist.

### 6. 🟢 Minor: `conftest.py` only patches `migrations`, not the other new modules

The test fixture patches path constants on `(app_target, app_helpers, story_io, migrations, event_db, lore_db, state_db)`. However, `debug_helpers.py`, `gm_plan.py`, and `state_updates.py` don't define any path constants directly — they import from `story_io`. This is correct and no action needed; just confirming the approach is sound.

---

## 📝 Verdict

**Overall: Approve with minor suggestions.** The refactoring is clean and well-scoped. The main actionable item is de-duplicating `_parse_env_int` (#1). The circular import patterns (#3, #4) are pre-existing and acceptable in the short term but should be tracked for future cleanup.

lainn9527:
Checked the review notes and I'm not making follow-up changes in this PR.

- `_parse_env_int` duplication is intentional here to keep the extracted modules independent from `app_helpers` and avoid reintroducing cross-module config coupling.
- `_DynamicPath` is only used at path-consumer call sites (`os.path.*`, `open`, `shutil`), so the current surface is acceptable for the migration use case.
- The deferred `state_updates -> gm_pipeline` and `migrations -> app_helpers` imports are known technical debt, but they are pre-existing dependency patterns rather than regressions from this refactor.
- `save_parsed()` call-site expansion in `core_routes.py` is deliberate and currently scoped to the migration/init path; I checked for other callers before changing it.

I'm treating those as follow-up cleanup items rather than blockers for this refactor.

PR #161, opened 2026-03-07
Move internal Python modules under story_core

lainn9527:
## Summary
- move root-level internal Python modules into the `story_core/` package
- keep `app.py` as the Flask entrypoint and add a thin `auto_play.py` wrapper for CLI compatibility
- update imports, scripts, tests, and docs to use the new package layout

## Validation
- python3 -m compileall app.py auto_play.py story_core routes scripts tests
- pytest


lainn9527:
## Review Summary

PR #165, opened 2026-03-07
Add ACTIONS marker for GM option stripping

lainn9527:
## Summary
- add an `<!--ACTIONS-->` marker contract for GM option blocks and strip those blocks before recent-context reuse, compaction, and async extraction
- keep the regex fallback, but only strip numbered choices when they follow a `可選行動` or `你打算` heading
- hide the new marker in the frontend and add regression tests for helper, compaction, and GM response processing

## Testing
- pytest tests/test_fate_strip.py tests/test_compaction.py tests/test_extract_tags_async.py tests/test_tag_extraction.py
- pytest tests/test_context_injection.py


lainn9527 (commented):
## Review

Core logic is correct and well-tested. The two-pronged approach (marker-based primary + tightened regex fallback) is a clean solution to the false-positive stripping bug.

### Issues to clarify

1. **`markdownToHtml()` double-strips** (`app.js:2510`): `stripHiddenTags()` is already called in `renderMessage()` before `markdownToHtml()`. Is there a code path where `markdownToHtml()` is called without prior `stripHiddenTags()`? If not, the addition is redundant.

2. **System prompt instruction duplication**: The `<!--ACTIONS-->` instruction is added to both `story_core/prompts.py` and `story_design/story_original/system_prompt.txt`. If both end up in the final composed prompt, the GM sees the instruction twice.

3. **Minor style**: `_ACTIONS_BLOCK_RE` uses `(?:.|\n)*` — rest of codebase uses `[\s\S]*` for the same purpose. Not blocking.

4. **`re.DOTALL` removed from `_CHOICE_BLOCK_RE`**: Since the pattern uses `[^\n]` instead of bare `.`, removing it has no functional impact — just confirming this was intentional.

None of these are blocking. LGTM with the above clarifications.

lainn9527:
Checked these points:

1. `markdownToHtml()` double-strip: intentional. There are multiple call sites that pass raw text directly into `markdownToHtml()` (`msg.content`, `gmMsg.content`, debug instruction text, etc.), so moving hidden-tag stripping into the helper makes the new `<!--ACTIONS-->` marker safe across all render paths instead of relying on each caller to remember to pre-strip. The existing explicit `stripHiddenTags()` calls on streaming paths are now redundant-but-harmless.

2. Prompt duplication: not happening in the composed GM prompt. `_build_story_system_prompt()` loads `story_design/<story_id>/system_prompt.txt` if it exists and only falls back to `story_core/prompts.py` when the design prompt is missing, so the GM only sees one source of the `<!--ACTIONS-->` instruction.

3. `_ACTIONS_BLOCK_RE` style: agreed this is just style. I left it as-is for now since it is covered by the new helper tests and the review was non-blocking, but I can normalize it to `[\s\S]*` in a follow-up if we want strict consistency.

4. `re.DOTALL` removal: yes, intentional. After tightening the fallback regex to line-bounded `[^\n]+`, `DOTALL` has no effect anymore.

So from my side there is no functional follow-up needed for this review.

